PR #7, opened 2020-02-28
ensure locals have consistent hashes to prevent redundent data transport

coryb:
Currently when reusing a locals definition I will see something like:
```
 => local://5ynzg3qmv2r4js718y3gkrorc                     0.1s
 => => transferring 5ynzg3qmv2r4js718y3gkrorc: 23.55kB    0.1s
 => local://63vy01nkeziqyd4sxvont84r1                     0.2s
 => => transferring 63vy01nkeziqyd4sxvont84r1: 23.55kB    0.1s
```

With this patch it will consistently hash based on path and local options:
```
 => local://bcb9151403a8f8668bd002cd4dcae9126ba9ae7910ce0465a66998fda3254ce3                0.1s
 => => transferring bcb9151403a8f8668bd002cd4dcae9126ba9ae7910ce0465a66998fda3254ce3: 514B  0.1s
```

### DIFF
--- a/codegen/codegen.go
+++ b/codegen/codegen.go
@@ -9,7 +9,6 @@ import (
 	shellquote "github.com/kballard/go-shellquote"
 	"github.com/moby/buildkit/client/llb"
 	"github.com/moby/buildkit/client/llb/imagemetaresolver"
-	"github.com/moby/buildkit/identity"
 	"github.com/moby/buildkit/solver/pb"
 	digest "github.com/opencontainers/go-digest"
 	"github.com/openllb/hlb/ast"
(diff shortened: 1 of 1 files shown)


codegen/codegen.go
```diff
@@ -300,15 +300,18 @@ func emitFilesystemSourceStmt(info *CodeGenInfo, scope *ast.Scope, call *ast.Cal
...
-
 		var opts []llb.LocalOption
 		for _, iopt := range iopts {
 			opt := iopt.(llb.LocalOption)
 			opts = append(opts, opt)
 		}
 
+		// get a consistent hash for this local (path + options) so we don't
+		// transport the same content multiple times when referenced repeatedly
+		sum := sha256.Sum256([]byte(fmt.Sprintf("%s:%#v", path, opts)))
```
coryb: I will keep working on this, the salt string ends up looking like: `.:[]llb.LocalOption{(llb.localOptionFunc)(0x1a2cae0)}` which does not seem like a great value to hash.

hinshun: Yeah, you could try:
```go
st = llb.Image(id, opts...)

dgst, _, _, err := st.Output().Vertex().Marshal(&llb.Constraints{})
if err != nil {
    return st, err
}

return st, nil
```

hinshun: Hm.. yeah, kind of a chicken and egg problem here. The `llb.Local` gives a `llb.State` that we can get a digest, but we also need it to create the `llb.Local`.

codegen/codegen.go
```diff
@@ -300,15 +299,27 @@ func emitFilesystemSourceStmt(info *CodeGenInfo, scope *ast.Scope, call *ast.Cal
...
 		for _, iopt := range iopts {
 			opt := iopt.(llb.LocalOption)
 			opts = append(opts, opt)
 		}
 
+		// get a consistent hash for this local (path + options) so we don't
+		// transport the same content multiple times when referenced repeatedly
+
+		// first get serialized bytes for this llb.Local state
+		tmpSt := llb.Local("constant-id-for-consistent-hashing", opts...)
```
hinshun: Could we just use `""`?

coryb: yeah, in theory, I suspect `""` is an invalid value, and was concerned at some point in the future there might be a validation error when we try to marshal it.

Issue #16, opened 2020-03-04
Add minimal travis yml


hinshun:
We don't have a great test suite atm, but we should have CI to at least make sure the code we're merging continues to compile.

PR #19, opened 2020-03-05
[#16] ensure code compiles and passes minimal linters

coryb:


coryb:
Okay, using github actions instead of travis now, seems to work.  The hlb + go mod stuff is pretty slow, we might be able to optimize that.  This is at least showing that we have a bunch of lint errors that should be cleaned up.  The workflow is already on master (had to write there to get the PR build to run it).

hinshun:
Awesome!

PR #23, opened 2020-03-12
update for linting, remove unused code

coryb:


source.hlb
```diff
@@ -59,7 +59,7 @@ fs goLint() {
 }
 
 fs goBuild(string package) {
-	image "golang:1.12-alpine"
```
hinshun: Do we need to update `go.mod`'s go version too?

coryb: I think that woudl break the crosscompile via `dockercore/golang-cross:1.12.5` so I think we need to wait for that to be available with 1.14 or find a replacement for the golang-cross image.

Issue #26, opened 2020-03-14
hlb parse error hidden in output


coryb:
When testing errors in parsing, I noticed when the log-output is tty, the error message is masked:
```
$ ./hlb --addr docker-container://buildkitd run ./tests/bug.hlb
[+] Building 0.2s (1/1) FINISHED
 => ERROR compiling ./tests/bug.hlb                                                                                                                                                 0.0s
------
 > compiling ./tests/bug.hlb:
------
```

But we can see it when `log-output` is `plain`:
```
$ ./hlb --addr docker-container://buildkitd run --log-output plain ./tests/bug.hlb
#1 compiling ./tests/bug.hlb
#1 ERROR: ./tests/bug.hlb:3:3: unexpected filesystem statement: "nothing"
------
 > compiling ./tests/bug.hlb:
------
```

hinshun:
@coryb Did moving #9 implementation to `checker` fix this issue?

coryb:
I suspect this will recur with any errors from the checker, but I can't find a way to reproduce currently so I guess we can close this for now.

Issue #9, opened 2020-02-28
Seg fault on some inputs


quidryan:
This produces a seg fault: 

```
fs default() {
  scratch
  nothing fs { scratch; }
}
fs nothing(fs repo) {
    scratch
}
```

quidryan:
Sample output:
```
➜  managed-artifacts-generator-baseos git:(master) ✗ hlb run syntaxerror.hlb
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x478b48c]
goroutine 1 [running]:
github.com/openllb/hlb/codegen.emitChainStmt.func1(0x4982ec0, 0xc000325d40, 0x4993ecf, 0x2)
        /go/src/hlb/codegen/codegen.go:165 +0xdc
github.com/openllb/hlb/codegen.emitBlock(0xc0003ef250, 0xc00033ca80, 0x4993ecf, 0x2, 0xc0003ef2d0, 0x2, 0x2, 0x49cdaf8, 0x8, 0xc0004f5578, ...)
        /go/src/hlb/codegen/codegen.go:146 +0x529
github.com/openllb/hlb/codegen.emitFilesystemBlock(0xc0003ef250, 0xc00033ca80, 0xc0003ef2d0, 0x2, 0x2, 0x49cdaf8, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/hlb/codegen/codegen.go:185 +0xbd
github.com/openllb/hlb/codegen.emitFuncDecl(0xc0003ef250, 0xc00033ca20, 0xc000372360, 0xc000318700, 0x0, 0x0, 0x49cdaf8, 0xc00041ccd0, 0xc0004f5808, 0x40df6fb, ...)
        /go/src/hlb/codegen/decl.go:44 +0x717
github.com/openllb/hlb/codegen.emitFilesystemFuncDecl(0xc0003ef250, 0xc00033ca20, 0xc000372360, 0xc000318700, 0x49cdaf8, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/hlb/codegen/decl.go:55 +0xa0
github.com/openllb/hlb/codegen.Generate(0xc000318700, 0xc00033ca00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /go/src/hlb/codegen/codegen.go:52 +0x4c0
[...]

PR #27, opened 2020-03-14
[#9] prevent panic on unexpected filesystem statement

coryb:
This is likely better addressed when we add the ability to create fs methods, but for now this will bail on invalid statements when processing filesystems.

So given:
```
fs default() {
  scratch
  nothing fs { scratch; }
}
fs nothing(fs repo) {
    scratch
}
```

We will now see:
```
./hlb --addr docker-container://buildkitd run --log-output plain ./tests/bug.hlb
#1 compiling ./tests/bug.hlb
#1 ERROR: ./tests/bug.hlb:3:3: unexpected filesystem statement: "nothing"
------
 > compiling ./tests/bug.hlb:
------
```

I did file #26 since the error message is now shown when with tty output.

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -322,65 +322,64 @@ func (c *checker) checkBlockStmt(scope *parser.Scope, typ parser.ObjType, block 
 			panic("implementation error")
 		}
 
-		lookupByType := builtin.Lookup.ByType[typ]
-		if !foundSource {
-			// If the function is not a builtin, retrieve it from the scope and then
-			// type check it.
-			_, ok := lookupByType.Func[name]
-			if !ok {
-				obj := scope.Lookup(name)
-				if obj == nil {
-					return ErrFirstSource{call}
-				}
+		// If the function is not a builtin, retrieve it from the scope and then
+		// type check it.
+		funcLookup, ok := builtin.Lookup.ByType[typ].Func[name]
+		if ok {
+			// If any other statement after the first is a source statement, then error.
+			//
+			// Consider the following:
+			//
+			// ```hlb
+			// fs default() {
+			//     image "alpine"
+			//     image "busybox"
+			// }
+			// ```
+			//
+			// Redeclaring a new source makes the previous instructions orphaned from
+			// the graph.
+			if foundSource && funcLookup.IsSource {
+				return ErrOnlyFirstSource{call}
+			}
+		} else {
+			obj := scope.Lookup(name)
+			if obj == nil {
+				return ErrIdentNotDefined{Ident: call.Func.Ident}
+			}
 
-				// The retrieved object may be either a function declaration or a field
-				// in the current scope's function parameters.
-				var callType *parser.Type
-				switch obj.Kind {
-				case parser.DeclKind:
-					switch n := obj.Node.(type) {
-					case *parser.FuncDecl:
-						callType = n.Type
-					case *parser.AliasDecl:
-						callType = n.Func.Type
+			// The retrieved object may be either a function declaration or a field
+			// in the current scope's function parameters.
+			var callType *parser.Type
+			switch obj.Kind {
+			case parser.DeclKind:
+				switch n := obj.Node.(type) {
+				case *parser.FuncDecl:
+					callType = n.Type
+					// if we already have a source and this function is not a method
+					// so it will be returning an invalid second source
+					if foundSource && n.Method == nil && callType.ObjType == typ {
+						return ErrOnlyFirstSource{CallStmt: call}
 					}
-				case parser.FieldKind:
-					field, ok := obj.Node.(*parser.Field)
-					if ok {
-						callType = field.Type
-					}
+				case *parser.AliasDecl:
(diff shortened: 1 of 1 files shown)


codegen/codegen.go
```diff
@@ -666,6 +666,8 @@ func (cg *CodeGen) EmitFilesystemChainStmt(ctx context.Context, scope *parser.Sc
 				llb.Copy(input, src, dest, opts...),
 			)
 		}
+	default:
+		return nil, fmt.Errorf("%s unexpected filesystem statement: %q", checker.FormatPos(call.Pos), call.Func.Ident.Name)
```
hinshun: This should be done in the `checker` package rather than during code generation. We want to fail earlier, because in the future (with multi-targets, etc) we might have ran a few solves at this point.

coryb: Okay, I am trying to understand the checker structure/logic now.

coryb:
Updated PR to move validation to the checker.  Most of the diff is whitespaces changes as I de-dented a block, so you might want to check the ignore whitespace option.   Now you will see this error:
```
$ ./hlb --addr docker-container://buildkitd run --log-output plain ./tests/bug.hlb
./tests/bug.hlb:3:3: only first statement must be source
```

The code will have to be modified when/if we allow for `fs (fs)` methods, but for now this will prevent multiple sources statements in a single `fs` block.

checker/checker.go
```diff
@@ -323,46 +323,46 @@ func (c *checker) checkBlockStmt(scope *parser.Scope, typ parser.ObjType, block
...
-						callType = field.Type
-					}
+			// The retrieved object may be either a function declaration or a field
+			// in the current scope's function parameters.
+			var callType *parser.Type
+			switch obj.Kind {
+			case parser.DeclKind:
+				switch n := obj.Node.(type) {
+				case *parser.FuncDecl:
+					callType = n.Type
```
hinshun: I think even though we don't support user defined methods atm, we can implement checking for function/method already.

checker/checker.go
```diff
@@ -323,46 +323,46 @@ func (c *checker) checkBlockStmt(scope *parser.Scope, typ parser.ObjType, block
...
-			_, ok := lookupByType.Func[name]
-			if !ok {
-				obj := scope.Lookup(name)
-				if obj == nil {
-					return ErrFirstSource{call}
-				}
+		// If the function is not a builtin, retrieve it from the scope and then
+		// type check it.
+		_, ok := lookupByType.Func[name]
+		if !ok {
```
hinshun: I think we can make this implementation more clean. We should do branch off here for:
```go
		_, ok := lookupByType.Func[name]
		if ok {
			// Handle builtins
			// lookupByType.Func[name].IsSource ...
		} else {
			// Handle user-defined
		}

		// Setting some shared variable between the two branches, we
		// abstract away builtin vs. user-defined nad check ErrOnlyFirstSource here.
```

checker/checker.go
```diff
@@ -378,10 +378,12 @@ func (c *checker) checkBlockStmt(scope *parser.Scope, typ parser.ObjType, block
 		//
 		// Redeclaring a new source makes the previous instructions orphaned from
 		// the graph.
-		if lookupByType.Func[name].IsSource {
+		if foundSource && lookupByType.Func[name].IsSource {
 			return ErrOnlyFirstSource{call}
```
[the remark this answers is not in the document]
coryb: I suspect the linter allows for inline struct usage when the struct only has 1 member.

checker/checker.go
```diff
@@ -322,65 +322,64 @@ func (c *checker) checkBlockStmt(scope *parser.Scope, typ parser.ObjType, block
...
+			// in the current scope's function parameters.
+			var callType *parser.Type
+			switch obj.Kind {
+			case parser.DeclKind:
+				switch n := obj.Node.(type) {
+				case *parser.FuncDecl:
+					callType = n.Type
+					// if we already have a source and this function is not a method
+					// so it will be returning an invalid second source
+					if foundSource && n.Method == nil && callType.ObjType == typ {
```
hinshun: Just calling it out here, I think it makes sense that the typing is higher priority than it being a source or not. :+1:

checker/checker.go
```diff
@@ -322,65 +322,64 @@ func (c *checker) checkBlockStmt(scope *parser.Scope, typ parser.ObjType, block
...
-			// type check it.
-			_, ok := lookupByType.Func[name]
-			if !ok {
-				obj := scope.Lookup(name)
-				if obj == nil {
-					return ErrFirstSource{call}
-				}
+		// If the function is not a builtin, retrieve it from the scope and then
+		// type check it.
+		funcLookup, ok := builtin.Lookup.ByType[typ].Func[name]
```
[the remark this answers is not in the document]
coryb: This appears safe since we are not dealing with any pointer types, I changed the code to this and it ran without exceptions:
```
funcLookup, ok := builtin.Lookup.ByType["foobar"].Func[name]
if ok {
```

Issue #10, opened 2020-02-28
Add `--version/--v` flag to show version.


hinshun:
We should add a package `version` with some golang constants that we replace at build time using ldflags.

Similar to: https://github.com/containerd/containerd/blob/master/version/version.go

coryb:
resolved via #28

PR #28, opened 2020-03-16
add version command and infer version from git tag

coryb:
Fixes issues #10 

This uses the git tag + distance to version the binary.  It adds an `hlb version` command which will print something like:
```
$ build/hlb-darwin-amd64 version
0.1-43-gf1453b1
```
The `0.1` is from the version tag, the `43` is the distances in commits since the version tag, and the `gf1453b1` is a `g` plus the git sha for HEAD.  The version string is derived from `git describe --tags --always`.  The version will have an extra `.dirty` on the end if it was compiled from modified sources.

I am not happy with the changes necessary to the `go.hlb` file, but I could not figure out anything better.   At least it seems to work.

### DIFF
--- a/go.hlb
+++ b/go.hlb
@@ -4,25 +4,31 @@ export crossBinaries
 
 export lint
 
-fs build(fs src, string package) {
+string versionCmd() {
+	format "git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'"
+}
+
+fs build(fs src, string package, string verPackage) {
 	image "golang:1.14-alpine"
 	run "apk add -U git gcc libc-dev"
 	env "GO111MODULE" "on"
 	dir "/go/src/hlb"
 	run string {
-		format "/usr/local/go/bin/go build -o /out/binary -ldflags '-linkmode external -extldflags -static' -a %s" package
+		format "v=$(%s) && /usr/local/go/bin/go build -o /out/binary -ldflags \"-linkmode external -extldflags -static -X %s.Version=$v\" -a %s" string { versionCmd; }  package verPackage
 	} with option {
 		cacheMounts src
 		mount fs { scratch; } "/out" as binary
 	}
 }
 
-fs crossBuild(fs src, string package) {
+fs crossBuild(fs src, string package, string verPackage) {
 	image "dockercore/golang-cross:1.12.5" with option { resolve; }
 	env "GOPATH" "/root/go"
 	env "GO111MODULE" "on"
 	dir "/go/src/hlb"
-	run "/cross/build" package with option {
+	run string {
+		format "v=$(%s) && LDFLAGS=\"-X %s.Version=$v\" /cross/build %q" string { versionCmd; } verPackage package 
+	} with option {
 		cacheMounts src
 		mount fs { git "https://github.com/hinshun/go-cross.git" ""; } "/cross" with option {
 			sourcePath "/scripts"
--- a/cmd/hlb/command/app.go
+++ b/cmd/hlb/command/app.go
@@ -30,6 +30,7 @@ func App() *cli.App {
 	}
 
 	app.Commands = []*cli.Command{
+		versionCommand,
 		runCommand,
 		formatCommand,
 		moduleCommand,
--- a/cmd/hlb/command/version.go
+++ b/cmd/hlb/command/version.go
@@ -1,0 +1,17 @@
+package command
+
+import (
+	"fmt"
+
+	"github.com/openllb/hlb"
+	cli "github.com/urfave/cli/v2"
+)
+
+var versionCommand = &cli.Command{
+	Name:  "version",
+	Usage: "prints hlb tool version",
+	Action: func(c *cli.Context) error {
+		fmt.Println(hlb.Version)
+		return nil
+	},
+}
--- a/source.hlb
+++ b/source.hlb
@@ -5,7 +5,7 @@ fs default() {
 }
 
 fs crossHLB() {
-	go.crossBinaries src "github.com/openllb/hlb/cmd/hlb"
+	go.crossBinaries src "github.com/openllb/hlb/cmd/hlb" "github.com/openllb/hlb"
 }
 
 fs lint() {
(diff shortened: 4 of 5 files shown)


hinshun:
@coryb What do you think we can add to make this more natural?

coryb:
@hinshun not entirely sure. I would like to move the `git` usage out into `source.hlb` rather than in the `go.hlb` library.  To do that, I think we would need to support generic ldflags arguments to the builder.

I was thinking something like 
```
fs build(fs src, string package, variadic string ldflags) {
}
```
But I could not figure out how to convert the variadic string into a single string that we could pass to `-ldflags` option. I guess we could just take a single argument and append that value to the defaults like:
```
fs build(fs src, string package, string ldflags) {
  run string {
    format "/usr/local/go/bin/go build -o /out/binary -ldflags \"-linkmode external -extldflags -static %s\" -a %s" package ldflags
  }
}
```

The next issue was that I could not come up with an effective way to capture the output from:
```
git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'
```
I need to run that command with the local `.git` context and then capture that output into a variable to pass to the go.build routine. Any thoughts on how that would be possible?  Maybe something like:
```
fs gitVersionLookup() {
  local "."
  run " git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'" with option {
    stdout as myver
  }
}
fs goBuild() {
  go.build fs { local "."; } "github.com/openllb/hlb/cmd/hlb" string {
    format "-X github.com/openllb/hlb.Version=%s" myver
  }
}
```

An alternative to doing this in hlb would be to just allow for an ad-hoc version parameter that we can pass into the hlb command like:
```
./hlb run --target goBuild --param myver=$(git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//') ./source.hlb
```
But if we go that way, the next step is to create a Makefile to wrap that.

hinshun:
> 
> I was thinking something like
> 
> ```
> fs build(fs src, string package, variadic string ldflags) {
> }
> ```
> 
> But I could not figure out how to convert the variadic string into a single string that we could pass to `-ldflags` option. I guess we could just take a single argument and append that value to the defaults like:
> 
> ```
> fs build(fs src, string package, string ldflags) {
>   run string {
>     format "/usr/local/go/bin/go build -o /out/binary -ldflags \"-linkmode external -extldflags -static %s\" -a %s" package ldflags
>   }
> }
> ```

I think the choice between one string, and `variadic` is a matter of taste. But sounds like we should implement `string join(string sep, variadic string values)`!

> The next issue was that I could not come up with an effective way to capture the output from:
> 
> ```
> git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'
> ```
> 
> I need to run that command with the local `.git` context and then capture that output into a variable to pass to the go.build routine. Any thoughts on how that would be possible? Maybe something like:
> 
> ```
> fs gitVersionLookup() {
>   local "."
>   run " git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'" with option {
>     stdout as myver
>   }
> }
> fs goBuild() {
>   go.build fs { local "."; } "github.com/openllb/hlb/cmd/hlb" string {
>     format "-X github.com/openllb/hlb.Version=%s" myver
>   }
> }
> ```

 I rather not conflate `local "."` (which means making a copy of your local directory starting at ".") with running things locally. Perhaps a string function like: `string localrun(variadic string args)` that functions like the `run` command except it's local.

> An alternative to doing this in hlb would be to just allow for an ad-hoc version parameter that we can pass into the hlb command like:
> 
> ```
> ./hlb run --target goBuild --param myver=$(git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//') ./source.hlb
> ```
> 
> But if we go that way, the next step is to create a Makefile to wrap that.

Yes, and this is already possible with `echo | hlb run`. I'm still thinking of locally run stdout belongs in hlb... Perhaps you could sync up the `.git` into a container to run the command, and then somehow access the stdout in a `string` variable?

go.hlb
```diff
@@ -4,25 +4,31 @@ export crossBinaries
...
+	format "git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'"
+}
+
+fs build(fs src, string package, string verPackage) {
 	image "golang:1.14-alpine"
 	run "apk add -U git gcc libc-dev"
 	env "GO111MODULE" "on"
 	dir "/go/src/hlb"
 	run string {
-		format "/usr/local/go/bin/go build -o /out/binary -ldflags '-linkmode external -extldflags -static' -a %s" package
```
hinshun: I think one feature we should implement is if the function has 0 arguments, you should be able to call it without a `string` func literal.

hinshun: Created design issue: #30

coryb:
> > ```
> > fs gitVersionLookup() {
> >   local "."
> >   run " git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'" with option {
> >     stdout as myver
> >   }
> > }
> > fs goBuild() {
> >   go.build fs { local "."; } "github.com/openllb/hlb/cmd/hlb" string {
> >     format "-X github.com/openllb/hlb.Version=%s" myver
> >   }
> > }
> > ```
> 
> I rather not conflate `local "."` (which means making a copy of your local directory starting at ".") with running things locally. Perhaps a string function like: `string localrun(variadic string args)` that functions like the `run` command except it's local.

I was using `local "."` just for illustration.  I don't think adding `localrun` would be a good idea, at least not at this point. I really want to execute a command on a provided `fs` (`local`, `git`, `image`, `http` etc) and capture the stdout.   I am not sure how to represent the concept of capturing stdout to a string.  I mocked `with option { stdout as myver; }` which is not great since I can't see any other practical reason to have a `stdout` option.

When building `run` commands I frequently find myself doing this pattern`x=$(...) && command --option=$x` which makes it pretty incoherent.  I feel like we might need a higher lever concept for this sort of usage, not sure how possible it is with llb though.

PR #29, opened 2020-03-16
[#12] add support for `auto` log-output option

### LINKED ISSUE #12: Add `auto` log-output for non-console environments.
Some environments like Jenkins, doesn't have a tty available. We need an `auto` log-outputter like BuildKit, which will switch between `tty` and `plain` depending if a tty is available on stdin.

coryb:
fixes #12.  If LogOutput is unset or set to `auto` we will detect if the Output is a tty and choose between `plain` and `tty` for the LogOutput setting.

### DIFF
--- a/cmd/hlb/command/run.go
+++ b/cmd/hlb/command/run.go
@@ -7,6 +7,7 @@ import (
 	"os"
 
 	"github.com/docker/buildx/util/progress"
+	"github.com/mattn/go-isatty"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/util/appcontext"
 	"github.com/openllb/hlb"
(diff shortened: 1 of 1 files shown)


cmd/hlb/command/run.go
```diff
@@ -102,6 +103,16 @@ func Run(ctx context.Context, cln *client.Client, rc io.ReadCloser, opts RunOpti
 	}
 
 	var progressOpts []solver.ProgressOption
+	if opts.LogOutput == "" || opts.LogOutput == "auto" {
+		// asusme plain interface, will upgrade if we detect tty
```
hinshun: s/asusme/Assume/

coryb: fixed.

PR #49, opened 2020-03-26
Implement multi-outputs for HLB

### LINKED ISSUE #18: Defining outputs in HLB rather than in command-line
When we land multi-targets in #47, it is only helpful if we can output the targets to different locations.

1. Output as soon as targets have been solved, without waiting for other parts of the graph
2. Multiple outputs per target, useful when wanting to push to multiple registries

Proposal
---

## HLB CLI

We'll move away from `hlb run --download <dest> ` / `hlb run --push <ref>` and instead bring it down into `hlb` functions.

But you'll still be able to specify output flags on the command line:
- `hlb run --target foo,download=.`

The target is splitted on the sentinel `,` and the first value is always the target name, followed by key values for output flags.

## Builtin Functions
```hlb
# Outputs the filesystem to external sources. If no options are
# defined, this is a no-op.
#
# @return the filesystem unchanged.
fs (fs) output()

# Pushes the filesystem to a registry following the distribution
# spec: https://github.com/opencontainers/distribution-spec/
#
# @param ref a docker registry reference. if not fully qualified, it will be
# expanded the same as the docker CLI.
# @return an option to push the filesystem to a registry.
option::output push(string ref)

# Downloads the filesystem to a local path.
#
# @param localPath the destination filepath for the filesystem contents.
# @return an option to download a filesystem to the local system.
option::output download(string localPath)

fs default() {
    scratch
    mkfile "/out" 0o644 "hello world"
    output with option {
        push "us-east-1.registry/foo:latest"
        push "us-west-2.registry/foo:latest"
    }
}
```

By running the `default` target, we end up solving the graph and pushing it to two different registries.

**Note this changes HLB from a compiler to more of an interpreter, unless BuildKit supports these output vertices in LLB**

## Outputs happen concurrently with rest of graph

Output returns the same `input` filesystem so you can continue chaining statements afterwards:
```hlb
fs default() {
    scratch
    mkfile "/out" 0o644 "hello world"
    output with option {
        push "foo:latest"
    }
    env "key" "value"
}
```

hinshun:
Fixes #18 

Allows outputs to be defined in a `fs` function.

Downloading to local directory:
```hlb
fs foo() {
	scratch
	mkfile "foo" 0o644 "hello foo"
	output with option {
		download "./foo"
	}
}
```

Pushing to multiple registries:
```hlb
fs foo() {
	scratch
	mkfile "foo" 0o644 "hello foo"
	output with option {
		dockerPush "us-east-1.myregistry.com/hinshun/foo"
		dockerPush "us-west-2.myregistry.com/hinshun/foo"
	}
}
```

Loading to local docker API:
```hlb
fs foo() {
	scratch
	mkfile "foo" 0o644 "hello foo"
	output with option {
		dockerLoad "foo:latest"
	}
}
```

You can also define commands after `output` but `output` will only output the snapshot of the filesystem at the line it was invoked:

```hlb
fs foo() {
	scratch
	mkfile "foo" 0o644 "hello foo"
	output with option {
		dockerLoad "foo:latest"
	}
	# File "bar" will not be in "foo:latest" exported to host.
	mkfile "bar" 0o644 "hello bar"
}
```


frontend.go
```diff
@@ -137,17 +137,19 @@ func Frontend(ctx context.Context, c client.Client) (*client.Result, error) {
 		return nil, err
 	}
 
-	st, err := cg.Generate(ctx, call, file)
+	_, err = cg.Generate(ctx, mod, []*parser.CallStmt{call})
```
coryb: I dont understand this code, why are we calling Generate and then throwing away the solve.Request? What is the point?

module/resolve.go
```diff
@@ -195,10 +195,10 @@ func (r *remoteResolver) Resolve(ctx context.Context, scope *parser.Scope, decl
 		return nil, err
 	}
 
-	var solveOpts []solver.SolveOption
-	for id, path := range cg.Locals {
-		solveOpts = append(solveOpts, solver.WithLocal(id, path))
-	}
+	// var solveOpts []solver.SolveOption
```
coryb: remove?

frontend.go
hinshun: I didn't want to spend too much time on `frontend.go` because its deprecated. It used to be a way to `hlb publish` a function, but we use libraries for that now. Maybe I should just delete this file.

module/resolve.go
hinshun: Note that `(*CodeGen).GenerateImport` still needs to return `llb.State` because we need to generate the vertex digest for the import.

hinshun: I addressed it here:
- https://github.com/openllb/hlb/pull/49/commits/668cb87ff69c5cc0500c82c89da53a894b3493c2#diff-5465bec959c3455c1a373bcf1569ad5a
- https://github.com/openllb/hlb/pull/49/commits/668cb87ff69c5cc0500c82c89da53a894b3493c2#diff-641e469b8b7432439f761283de3bfd0c

PR #54, opened 2020-03-29
[#51] remove methods and single-source requirements from fs blocks

### LINKED ISSUE #51: Remove methods from HLB
Problem
---

Currently, `fs` function like the one below:
```hlb
fs default() {
    image "alpine"
    run "echo hello"
}
```

Have the following restraints:
- `fs` body must have at least one statement.
- `fs` first statement must be a "source" statement (i.e. `git`, `image`, `http`, `local`, `scratch`).
- `fs` statements after the first must be methods. ("source" statements are not allowed).

Contrasted with `option` blocks:
- `option` body can be empty.
- `option` has no "source" statement, any statement is thought of as appending to the list of options.
- `option` block can invoke an `option` function, where the options are merged to the list of options.

When also taking #47 into consideration, `group` blocks are:
- `group` body can be empty.
- `group` also has no "source" statement, any statement is thought of as appending to a list of sequential targets.
- `group` block can invoke a `group` function, where the groups are merged to the list of sequential targets.

Having different rules for blocks depending on the return type is confusing, and the complexity also shows up in implementation of the compiler, as well as having to explain to users in documentation.

Proposal
---

I propose we unify these traits in a framework that can be applied to each block. In this proposal, methods are no longer a concept, you can think of every function as a "method" that modifies the value on the stack.

All blocks have the same rules:
- body can be empty
- conceptually, when running a function, the stack has one register which holds the zero-value of that type (`fs` is `llb.Scratch`, `option` is empty list of options). functions modify the value in this register and saves it to the register.

WIP (show examples)

coryb:
fixes #51 

This allows multiple sources to be used within a single fs.  fs "methods" will continue to operate on the previous state in the stack/chain.

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -271,13 +271,6 @@ func (c *checker) checkBlockStmt(scope *parser.Scope, typ parser.ObjType, block 
 		return c.checkOptionBlockStmt(scope, typ, block)
 	}
 
-	// Non-option blocks must specify a source statement.
-	if block.NumStmts() == 0 {
-		return ErrNoSource{block}
-	}
-
-	foundSource := false
-
 	for _, stmt := range block.NonEmptyStmts() {
 		if stmt.Bad != nil {
 			c.errs = append(c.errs, ErrBadParse{stmt})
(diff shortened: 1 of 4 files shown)


hinshun:
We'll need a follow up PR to remove the `(fs)` from all the `builtin.hlb`: https://github.com/openllb/hlb/blob/master/language/builtin.hlb#L123

PR #57, opened 2020-03-29
update builtin specs for recent changes

coryb:
Updates builtin.hlb for changes made in #54 and #56

### DIFF
--- a/gen/builtins.go
+++ b/gen/builtins.go
@@ -22,9 +22,8 @@ type BuiltinData struct {
 }
 
 type ParsedFunc struct {
-	Name     string
-	IsSource bool
-	Params   []*parser.Field
+	Name   string
+	Params []*parser.Field
 }
 
 func GenerateBuiltins(r io.Reader) ([]byte, error) {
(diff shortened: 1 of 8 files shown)


hinshun:
You'll also have to run `go generate ./gen` in order to update `builtin/builtin.go`.

We also want to remove `Method` from the CST! I can open a PR for that.

gen/builtins.go
```diff
@@ -41,7 +41,7 @@ func GenerateBuiltins(r io.Reader) ([]byte, error) {
 		}
 
 		isSource := false
-		if fun.Method == nil && fun.Type.Primary() != parser.Option {
+		if fun.Type.Primary() != parser.Option {
 			isSource = true
```
hinshun: We should remove `isSource` and `(FuncLookup).IsSource`.

gen/documentation.go
```diff
@@ -119,7 +119,6 @@ func GenerateDocumentation(r io.Reader) (*Documentation, error) {
 
 		funcDoc := &Func{
 			Type:   typ,
-			Method: fun.Method != nil,
```
hinshun: Also remove the field `(Func).Method` from the struct.

hinshun:
Lost inside the `Crosscompile` stage:
```
#5 53.10 go: error loading module requirements
```

PR #74, opened 2020-04-10
Buildkit 0.7.0-rc1

hinshun:
- Updating to BuildKit `0.7.0-rc1`
- `llb.Async` now available, many functions now require a `context.Context` and now returns error.
- Use `go 1.13.10` consistently across. `go 1.14` still has issues that many projects are staying on `go 1.13` until it's all resolved.
- Stays on fork on `nested-frontend`, PR #63 will unfork.


codegen/codegen.go
```diff
@@ -584,19 +599,6 @@ func (cg *CodeGen) EmitFilesystemChainStmt(ctx context.Context, scope *parser.Sc
...
-			stArg, err := cg.EmitStringExpr(ctx, scope, call, arg)
-			if err != nil {
-				return so, err
-			}
-			stArgs = append(stArgs, stArg)
-		}
-
-		so = func(st llb.State) (llb.State, error) {
-			return st.Args(stArgs...), nil
-		}
```
hinshun, line 599: `st.Args` was from the `netflix` fork. We don't have this in upstream yet.

PR #83, opened 2020-04-13
Asynchronously resolve image configs using gateway and marshal in parallel

hinshun:
- Moves marshalling/validation/get state metadata into goroutines to execute concurrently

```sh
$ hlb run -t foo -t bar -t baz source.hlb
[+] Building 0.9s (9/9) FINISHED
 => compiling [foo bar baz]                                                                                                                                           0.8s
 => resolve image config for docker.io/library/golang:alpine                                                                                                          0.5s
 => resolve image config for docker.io/library/busybox:latest                                                                                                         0.4s
 => resolve image config for docker.io/library/alpine:latest                                                                                                          0.6s
 => resolve image config for docker.io/library/node:alpine                                                                                                            0.6s
```

## TODO
- [ ] Still appears to be a panic from BuildKit `llb`. There are more unprotected maps there. We can either contribute more PRs upstream or think of how to deep copy `llb.State`.

### DIFF
--- a/codegen/codegen.go
+++ b/codegen/codegen.go
@@ -3,7 +3,6 @@ package codegen
 import (
 	"context"
 	"fmt"
-	"io"
 	"net"
 	"os"
 	"sort"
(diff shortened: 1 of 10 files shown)


hinshun:
This is far out of date.

PR #84, opened 2020-04-13
Add forward builtin to forward traffic from/to a local source

hinshun:
Run the example yourself:
```sh
hlb run -t nginx examples/forward.hlb
hlb run --log-output plain -t tcp examples/forward.hlb
```

```sh
$ hlb run --log-output plain -t tcp examples/forward.hlb
#1 compiling [tcp]
#1 DONE 1.1s
#2 docker-image://docker.io/library/alpine:latest
#2 resolve docker.io/library/alpine:latest
#2 resolve docker.io/library/alpine:latest 0.4s done
#2 DONE 0.4s
#3 /bin/sh -c apk add -U curl
#3 CACHED
#4 /bin/sh -c curl --unix-socket /nginx.sock http://localhost
#4 0.322   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
#4 0.322                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0<!DOCTYPE html>
#4 0.323 <html>
#4 0.323 <head>
#4 0.323 <title>Welcome to nginx!</title>
#4 0.323 <style>
#4 0.323     body {
#4 0.323         width: 35em;
#4 0.323         margin: 0 auto;
#4 0.323         font-family: Tahoma, Verdana, Arial, sans-serif;
#4 0.323     }
#4 0.323 </style>
#4 0.323 </head>
#4 0.323 <body>
#4 0.323 <h1>Welcome to nginx!</h1>
#4 0.323 <p>If you see this page, the nginx web server is successfully installed and
#4 0.323 working. Further configuration is required.</p>
#4 0.323
#4 0.323 <p>For online documentation and support please refer to
#4 0.323 <a href="http://nginx.org/">nginx.org</a>.<br/>
#4 0.323 Commercial support is available at
#4 0.323 <a href="http://nginx.com/">nginx.com</a>.</p>
#4 0.323
#4 0.323 <p><em>Thank you for using nginx.</em></p>
#4 0.323 </body>
#4 0.323 </html>
100   612  100   612    0     0   597k      0 --:--:-- --:--:-- --:--:--  597k
#4 DONE 0.5s
```


examples/forward.hlb
```diff
@@ -0,0 +1,47 @@
...
+	run "apk add -U openssh-client"
+	mkdir "/root/.ssh" 0o700
+	run "ssh-keyscan github.com >> /root/.ssh/known_hosts"
+	run "ssh -q -T git@github.com || true" with option {
+		ignoreCache
+		ssh
+	}
+}
+
+fs nginx() {
```
coryb, line 32: So I understand, the nginx target is running an nginx container on my localhost via shared docker.sock file?  If we run this target twice, it will be cached (but the daemon might not necessarily be running locally, perhaps I rebooted my laptop) and not restart it?
Not sure we have a strategy around that.  `ignoreCache` would make sure it is run again, but then if the container is already running locally that will also cause a failure.

I think the example is interesting, but it seems like there are some "sharp edges" here.

examples/forward.hlb
```diff
@@ -0,0 +1,47 @@
...
+	image "alpine" with option { resolve; }
```
coryb, line 33: minor nit, I dont think we need alpine here, I suspect the `dockerCli` base is sufficient since it is a static (no cgo) binary. Then you can change the `docker run` command to `./docker run ...`?

examples/forward.hlb
hinshun: Yes, we can definitely improve the actual command there so it will have smoother edges. Perhaps something like `docker rm -f nginx || true`.

examples/forward.hlb
hinshun: I think since we want to make the nginx function more friendly, we still want to have a shell (not necessarily alpine). Maybe busybox is a better choice here?

PR #85, opened 2020-04-14
Add codegen test framework

hinshun:
- Simplified how `llb.Local` id is calculated. The terminal op of the graph `def.Def[len(def.Def)-1]` is an empty vertex with an input to the last vertex's digest. Since that vertex also has its digests of its inputs and so on, the digest of the terminal op is a merkle hash for the graph.
- Change `solver/request` to be more testable
- Add `(solver.Request).Tree` to debug solve trees and compare results with expected value

```sh
❯ go test -v ./codegen
=== RUN   TestCodeGen
=== RUN   TestCodeGen/image
=== PAUSE TestCodeGen/image
=== RUN   TestCodeGen/call_function
=== PAUSE TestCodeGen/call_function
=== RUN   TestCodeGen/local
=== PAUSE TestCodeGen/local
=== CONT  TestCodeGen/image
=== CONT  TestCodeGen/local
=== CONT  TestCodeGen/call_function
--- PASS: TestCodeGen (0.00s)
    --- PASS: TestCodeGen/image (0.00s)
        codegen_test.go:153: .
            └── parallel
                └── single
                    └── [source]  identifier:"docker-image://docker.io/library/alpine:latest"

    --- PASS: TestCodeGen/local (0.00s)
        codegen_test.go:153: .
            └── parallel
                └── single
                    └── [source]  identifier:"local://sha256:9c47e3158363ec87074ac17b48adf8db980ffdc680a347d3106c9d8dc1ddebc5" attrs:<key:"local.session" value:"661ubccj9uggraywqw5imxova" >

    --- PASS: TestCodeGen/call_function (0.00s)
        codegen_test.go:153: .
            └── parallel
                └── single
                    └── [source]  identifier:"docker-image://docker.io/library/busybox:latest"

PASS
ok  	github.com/openllb/hlb/codegen	(cached)
```

### DIFF
--- a/codegen/codegen.go
+++ b/codegen/codegen.go
@@ -1849,6 +1849,10 @@ func (cg *CodeGen) LocalID(path string, opts ...llb.LocalOption) (string, error)
 		return "", err
 	}
 
+	// The terminal op of the graph def.Def[len(def.Def)-1] is an empty vertex with
+	// an input to the last vertex's digest. Since that vertex also has its digests
+	// of its inputs and so on, the digest of the terminal op is a merkle hash for
+	// the graph.
 	return digest.FromBytes(def.Def[len(def.Def)-1]).String(), nil
 }
 
--- a/codegen/codegen.go
+++ b/codegen/codegen.go
@@ -44,11 +44,11 @@ var (
 type StateOption func(llb.State) (llb.State, error)
 
 type CodeGen struct {
-	Debug   Debugger
-	request solver.Request
-	cln     *client.Client
+	Debug     Debugger
+	cln       *client.Client
+	sessionID string
 
-	localID         string
+	requests        []solver.Request
 	syncedDirByID   map[string]filesync.SyncedDir
 	fileSourceByID  map[string]secretsprovider.FileSource
 	agentConfigByID map[string]sockprovider.AgentConfig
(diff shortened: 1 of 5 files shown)


codegen/codegen.go
```diff
@@ -1818,17 +1840,20 @@ func (cg *CodeGen) EmitMountOptions(ctx context.Context, scope *parser.Scope, op
...
+
+	def, err := st.Marshal(context.Background())
 	if err != nil {
 		return "", err
 	}
 
-	// Next append the path so we have the path + options serialized hash input.
-	hashInput = append(hashInput, []byte(path)...)
-	return string(digest.FromBytes(hashInput)), nil
+	return digest.FromBytes(def.Def[len(def.Def)-1]).String(), nil
```
coryb, line 1852: Heh, this is incoherent, in your PR comment you added:
```
The terminal op of the graph def.Def[len(def.Def)-1] is an empty vertex with an input to the last vertex's digest. Since that vertex also has its digests of its inputs and so on, the digest of the terminal op is a merkle hash for the graph.
```
maybe a similar comment should be in code as well?

codegen/codegen_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/moby/buildkit/client/llb"
+	"github.com/openllb/hlb/checker"
+	"github.com/openllb/hlb/parser"
+	"github.com/openllb/hlb/solver"
+	"github.com/stretchr/testify/require"
+	"github.com/xlab/treeprint"
+)
+
+func Expect(st llb.State) solver.Request {
```
coryb, line 16: I think we might want to make this `func Expect(t *testing.T, st llb.State)` so that we can `require.NoError(t, err)` in the code. The panics will be hard to debug which test failed otherwise.

codegen/codegen_test.go
```diff
@@ -0,0 +1,159 @@
...
+	return solver.Single(&solver.Params{
+		Def: def,
+	})
+}
+
+type testCase struct {
+	name    string
+	targets []string
+	input   string
+	fn      func(cg *CodeGen) (solver.Request, error)
```
coryb, line 31: similar here, I think we will need to make this `func(t *testing.T, cg *CodeGen) solver.Request` so that we can simplify the error handling and pass the `t` to `Expect`.

PR #95, opened 2020-04-15
Test scope and identifiers

### LINKED ISSUE #86: Unit test bundle for scope and identifiers
Tests should go into `compile_test.go`, and the input program should be minimal as possible. There may be more than one test case per bullet point.

- [x] Function with duplicate names should produce error (functions, aliases).
- [x] Export name should exist in module scope.
- [x] Function with duplicate fields in the signature should produce error.
- [x] Calling selector that does not exist should produce error.
- [x] Calling selector on name that isn't an import should produce error.

antonio-osorio:
Resolves #86

### DIFF
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -11,7 +11,7 @@ import (
 type testCase struct {
 	name    string
 	input   string
-	errType interface{}
+	errType error
 }
 
 func cleanup(value string) string {
@@ -21,7 +21,7 @@ func cleanup(value string) string {
 	return result
 }
 
-func TestCompile(t *testing.T) {
+func TestChecker_Check(t *testing.T) {
 	for _, tc := range []testCase{{
 		"empty",
 		`
@@ -240,6 +240,15 @@ func TestCompile(t *testing.T) {
 		export myNonExistentFunction
 		`,
 		ErrIdentNotDefined{},
+	}, {
+		"errors when a selector is called on a name that isn't an import",
+		`
+		fs myFunction() {}
+		fs badSelectorCaller() {
+    		myFunction.build
+		}
+		`,
+		ErrNotImport{},
 	}} {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


checker/checker_test.go
```diff
@@ -183,6 +183,63 @@ func TestCompile(t *testing.T) {
 		}
 		`,
 		nil,
+	}, {
+		"errors with duplicate function names",
+		`
+		fs duplicateFunctionName() {}
+		fs duplicateFunctionName() {}
+		`,
+		ErrDuplicateDecls{},
```
hinshun, line 192: Can we modify how we test in order to ensure that the `err.Error()` matches what's returned by `checker.Check`?

antonio-osorio: Sounds good. I have updated the tests to have all the information needed to validate `err.Error()`. 

Do you think the position of the error is relevant? or is it ok to remove? Seems like the position would be very brittle and require re-skinning of test if, for example, the formatter would change.

checker/checker_test.go
```diff
@@ -183,28 +184,209 @@ func TestCompile(t *testing.T) {
...
+					Column:   7,
+				},
+				Name: "myFunction",
+			},
+		},
 	}} {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
+			//t.Parallel()
```
hinshun, line 301: Leftover?

checker/checker_test.go
```diff
@@ -21,7 +22,9 @@ func cleanup(value string) string {
 	return result
 }
 
-func TestCompile(t *testing.T) {
+func TestChecker_Check(t *testing.T) {
+	t.Parallel()
```
hinshun, line 26: `t.Parallel()` should be inside the `t.Run`

antonio-osorio: The whole test is about 10ms with most of the cases in under 1 ms. Having the parallel at the test case level mangles the output and makes it more difficult to debug. Specially since you cannot just run a specific case.

PR #99, opened 2020-04-16
Add localEnv string builtin

hinshun:
See test for example.

I chose `localEnv` to distinguish it from the `env` instruction for `fs`. Perhaps `getEnv` is better? WDYT?

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -424,6 +424,11 @@ var (
 							parser.NewField(parser.Str, "values", true),
 						},
 					},
+					"localEnv": FuncLookup{
+						Params: []*parser.Field{
+							parser.NewField(parser.Str, "key", false),
+						},
+					},
 				},
 			},
 		},
--- a/codegen/codegen.go
+++ b/codegen/codegen.go
@@ -414,6 +414,15 @@ func (cg *CodeGen) EmitStringChainStmt(ctx context.Context, scope *parser.Scope,
 		return func(_ string) (string, error) {
 			return fmt.Sprintf(formatStr, as...), nil
 		}, nil
+	case "localEnv":
+		key, err := cg.EmitStringExpr(ctx, scope, call, args[0])
+		if err != nil {
+			return nil, err
+		}
+
+		return func(_ string) (string, error) {
+			return os.Getenv(key), nil
+		}, nil
 	default:
 		// Must be a named reference.
 		obj := scope.Lookup(name)
--- a/codegen/codegen_test.go
+++ b/codegen/codegen_test.go
@@ -2,6 +2,7 @@ package codegen
 
 import (
 	"context"
+	"os"
 	"strings"
 	"testing"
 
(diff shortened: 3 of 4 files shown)


coryb:
Hmm, yeah, not a fan of the `localEnv` name, it works but feels cumbersome to me.

Was just thinking of a way we can solve this more generically. Maybe we add some `template` function like `format`? Then we could implement many string processing functions that way, or just embed [sprig](http://masterminds.github.io/sprig/).  Was thinking something like:
```
string home() {
  template '{{env "HOME"}}'
}

# pass optional arguments
string myJoin(variadic string parts) {
  template '{{. | join ","}}' parts
}
```

coryb:
Templating might also allow for conditionals in a constrained manor, used just with strings.  Ie something like:
```
string secretPath(string name) {
  template '{{if env "CI"}}/path/to/ci/secrets/{{{else}}/path/to/local/secrets{{end}}{{.}}' name
}
```
That is pretty hard to read, but quite functional.  Ideally we would have some heredoc syntax to allow that to more readable, but not sure if we want to go down that road yet.

coryb:
I guess I am leaning towards traditional heredoc syntax also, something like:
```
string secretDir() {
  template <<END
{{- if env "CI"- }}
  /path/to/ci/secrets/
{{- else- }}
  /path/to/local/secrets
{{ - end- }}'
END
}
```
or allow for an `<<-` option to trim indentation like:
```
string secretDir() {
  template <<-END
    {{- if env "CI"- }}
      /path/to/ci/secrets/
    {{- else- }}
      /path/to/local/secrets
    {{ - end- }}'
    END
}
```
but I suspect both will be a pain to implement in the parser.

slushie:
This reminds me of HCL string expansion, where lists have to be wrapped in strings which have a specific format. 😓 Personally, I'd prefer to take arguments from the command line as strings -- let the logic live outside of HLB, where its easy to conditionally produce any kind of value you need.

coryb:
I have run into several cases where I need string manipulation that is really build logic, rather than parameters.  One recent example, I was trying to figure out how to use the  appropriate crossbuild image for building Go binaries https://github.com/elastic/golang-crossbuild#build-tags
Roughly I need:
```
fs build() {
  image "<some image>"
  run "go build"
}
```
The `<some image>` is the hard part, ideally the users just pass in `"linux" and "arm64" as arguments, and we internally translate that to `docker.elastic.co/beats-dev/golang-crossbuild:1.14.2-arm`.  We could force the user to pass in the full image name, but nobody would use that.  With some limited templating we could at least generate the correct image.  Without templating I am looking at building a Go frontend just to do the if-else logic to generate the correct image name.

```
string buildImageTag(string os, string arch) {
  template <<END string { format "%s-%s" os arch }
{{if eq . "linux-amd64" }}
1.14.2-main
{{else if eq . "linux-arm64"}}
1.14.2-arm
{{else if eq . "..."}}
...
{{end}}
END
}
```

Issue #100, opened 2020-04-16
Call expressions as a shortcut for invoking functions


hinshun:
In the beginning, I designed HLB to invoke functions without parenthesis or commas like so:

```hlb
fs default() {
    image "alpine"
    env "key" "value"
    run "echo $key > /out/foo" with option {
        mount fs { scratch; } "/out" as fooOut
    }
}
```

Compare this to invoking with parens:
```hlb
fs default() {
    image("alpine")
    env("key", "value")
    run("echo $key > /out/foo") with option {
        mount fs { scratch(); } "/out" as fooOut
    }
}
```

It's a bit verbose and doesn't have the same readability as Dockerfiles so we ended up with what we have now. However, there are some common patterns that emerged:

### Running with string interpolated values
```hlb
run string {
    format "npm run %s" script
} with option {
    ...
}
```

### Mounting local and scratch filesystems
```hlb
run "go build -o /out/binary ./cmd/my-binary" with option {
    dir "/src"
    mount fs { local "."; } "/src"
    mount fs { scratch; } "/out"
}
```

### Proxying environment variables
```
option::run awsCredentials() {
    proxyEnv "AWS_ACCESS_KEY_ID"
    proxyEnv "AWS_SECRET_ACCESS_KEY"
    proxyEnv "AWS_SESSION_TOKEN"
}

option::run proxyEnv(string key) {
    env key string { localEnv key; }
}
```

Having to create a `fs { ... }` or `string { ... }` just to invoke one function is often too verbose. It only makes sense if you want to modify something so you end up having multiple statements:

```
run "make" with option {
    dir "/src"
    mount fs {
        local "."
        rm "./git/objects"
        mkdir "node_modules" 0o755
    } "/src"
}
```

I propose we support parens-style invocation of functions as an alternative, in order to reduce verbosity in these examples:

### Running with string interpolated values
```hlb
run format("npm run %s", script) with option {
    ...
}
```

### Mounting local and scratch filesystems
```hlb
run "go build -o /out/binary ./cmd/my-binary" with option {
    dir "/src"
    mount local(".") "/src"
    mount scratch "/out"
}
```

### Proxying environment variables
```
option::run awsCredentials() {
    proxyEnv "AWS_ACCESS_KEY_ID"
    proxyEnv "AWS_SECRET_ACCESS_KEY"
    proxyEnv "AWS_SESSION_TOKEN"
}

option::run proxyEnv(string key) {
    env key localEnv(key)
}
```

[...]

PR #102, opened 2020-04-17
Refactor codegen to allow calling functions without arguments as an expression

hinshun:
- Refactored `EmitChainXYZ` functions to `codegen/chain.go`
- Refactored `EmitChainXYZ` to depend on `expr *parser.Expr, args []*parser.Expr` instead of `call *parser.CallStmt` (So that `EmitXYZExpr` functions can call `EmitChainXYZ`. Reducing code duplication)
- Removed `call *parser.CallStmt` from `EmitStringExpr`, turns out it was unused.
- Made chain functions consistent (before: `StateOption`, `GroupChain`, `func(string) (string, error)`) (after: `FilesystemChain`, `GroupChain`, `StringChain`)

Now you can call builtin functions that have zero args directly:

```hlb
fs default() {
    image "alpine"
    run "touch /out/foo" with option {
        mount scratch "/out" as fooOut
    }
}
```

### DIFF
--- a/checker/checker.go
+++ b/checker/checker.go
@@ -482,6 +482,11 @@ func (c *checker) checkExpr(scope *parser.Scope, typ parser.ObjType, expr *parse
 }
 
 func (c *checker) checkIdentArg(scope *parser.Scope, typ parser.ObjType, ident *parser.Ident) error {
+	_, ok := builtin.Lookup.ByType[typ].Func[ident.Name]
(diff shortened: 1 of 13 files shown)


hinshun:
@coryb can you try compiling this PR and running it against `picardbot` to ensure I didn't break anything?

hinshun:
Phase 1 for #100

Issue #87, opened 2020-04-14
Unit test bundle for type checking and function signatures


hinshun:
Tests should go into compile_test.go, and the input program should be minimal as possible. There may be more than one test case per bullet point.

- [ ] Calling function that is wrong type for the function it is called in should produce error.
- [ ] Calling function with args that don't match its signature should produce error (length, basic literal type, function type, func lit type).
- [ ] Calling function with option that doesn't match the subtype should produce error.
- [ ] Correctly calling function with variadic field in its signature should succeed.

slushie:
#104 addresses these unit test, but some have been commented out because they currently fail on master.

Those tests highlight a few missing features in the checker and parser.

Checker currently doesn't identify incorrect types used as, or inside of, option blocks. That is `with wrongOptionType` and `with option { wrongOptionType; }` both pass, but shouldn't. It seems this also affects usage like `myfunc option::run { someCopyOpt; }` where the option block is passed as a value.

In discussion with @hinshun we realized the need for `CallSmt`s to be parsed as expressions, possibly as `CallExpr`. The parser and checker should also accept `BlockStmt`s that contain `BasicLit`s, so that `string myDir { "/var/lib/foo"; }` is a valid statement that assigns `"/var/lib/foo"` as the return value of `myDir`.

slushie:
```	// Option blocks may be empty and may refer to identifiers or function
	// literals that don't have a sub-type, so we check them differently.
	if strings.HasPrefix(string(typ), string(parser.Option)) {
		return c.checkOptionBlockStmt(scope, typ, block)
	}
``` in `checker.checkBlockStmt` seems to be the culprit for the checker issue.

@hinshun could you provide an example hlb program that has an option with a func literal that has no subtype? I'm not sure I know what that looks like and I want to write a test case for it so that I can try to fix the checker. NB if I comment out that block, most of the checks seem to work consistently.

**Comment by hinshun** [comment]
> @hinshun could you provide an example hlb program that has an option with a func literal that has no subtype? I'm not sure I know what that looks like and I want to write a test case for it so that I can try to fix the checker. NB if I comment out that block, most of the checks seem to work consistently.

I think what I meant by that comment is the following.

> Option blocks may be empty.

```hlb
fs default() {
    image "alpine"
    run "echo foo" with option {}
}
```

> may refer to identifiers

```hlb
fs default() {
    image "alpine"
    run "echo foo" with foo
}

option::run foo() {
    env "key" "value"
}
```

[...]

PR #104, opened 2020-04-17
Test function call semantics

slushie:
Addresses #87 

Some failing tests due to changes required in the `checker` package, provided in another PR.

### DIFF
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -387,7 +387,7 @@ func TestChecker_Check(t *testing.T) {
 				Found:    "int",
 			},
 		}, /*{
-			"func call with bad arg type: basic ident",
+			"func call with bad arg: basic ident",
 			`
 			fs default() {
 				myfunc s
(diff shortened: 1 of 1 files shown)


checker/checker_test.go
```diff
@@ -297,15 +305,229 @@ func TestChecker_Check(t *testing.T) {
...
+		fs default() {
+			myfunc "string"
+		}
+		fs myfunc(variadic option::run opts) {
+			image "busybox"
+			run "echo hi" with opts
+		}
+		`,
+		ErrWrongArgType{
+			Pos:      makePos(2, 8),
```
hinshun, line 320: Do you think we should abstract the testing away from these node positions? Or do you think they belong here?

checker/checker_test.go
```diff
@@ -297,15 +305,229 @@ func TestChecker_Check(t *testing.T) {
...
 			require.NoError(t, err)
 
-			err = Check(mod)
+			var r interface{}
+			func() {
+				defer func() {
+					r = recover()
+				}()
+				err = Check(mod)
+			}()
```
hinshun, line 529: We should probably return `errors.WithStack` instead of panicking in `Check`.

hinshun:
@slushie Needs a rebase.

PR #105, opened 2020-04-17
Resolve paths relative to the module's filepath

### LINKED ISSUE #2: String interpolation of environment variables
It would be nice to have environment variable substitution in Strings, especially in `local`. Environment variables are commonly used to put strings together or reference where files are on the local filesystem, like `$HOME`.

In this example, is impossible to generalize for different users without environment variables:

```
fs homeConfiguration() {
  local "/home/quidryan/.myconfig"
}
```

The used environment variables would probably have to be considered part of the input for caching.

hinshun:
- Paths are resolved to the module's filepath. So you can import modules and execute them without all the relative paths incorrect. See: https://github.com/hinshun/earthly/commit/d4ca01bda635eb24b3360ece9c9555e0707b9fdc
- Also supports home dir resolving via `go-homedir` (Fixes #2)
```hlb
fs homeConfiguration() {
  local "~/.myconfig"
}
```
- Supply `llb.SharedKeyHint` so that different relative strings that are pointing to the same underlying directory are deduplicated when transferring to BuildKit.

```hlb
# build.hlb
fs default() {
    local "."
}
```

```hlb
# subdir/build.hlb
fs default() {
    local "../"
}
```

### DIFF
--- a/codegen/codegen.go
+++ b/codegen/codegen.go
@@ -959,7 +959,7 @@ func (cg *CodeGen) EmitExecOptions(ctx context.Context, scope *parser.Scope, op 
 				)
 				switch srcUri.Scheme {
 				case "unix":
-					path = srcUri.Path
+					path = ResolvePathForNode(scope.Node, srcUri.Path)
 					id = digest.FromString(path).String()
 				default:
 					conn, err := net.Dial(srcUri.Scheme, srcUri.Host)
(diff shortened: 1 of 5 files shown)


codegen/codegen.go
```diff
@@ -1021,6 +1026,11 @@ func (cg *CodeGen) EmitExecOptions(ctx context.Context, scope *parser.Scope, op
 					return opts, err
 				}
 
+				localPath, err = ResolvePathForNode(scope.Node, localPath)
+				if err != nil {
+					return opts, err
+				}
```
coryb, line 1032: do we need to add SharedKeyHint for secrets as well?

hinshun: It's only a local option atm:
https://github.com/moby/buildkit/blob/4ae4bcd6a08e3a40aedead8d2edd571a11f75b55/client/llb/source.go#L345-L349

Issue #101, opened 2020-04-16
HEREDOC support


hinshun:
Multi-line string support is mainly in 2 forms:
1. Raw string literal like go.
2. HEREDOC has `<<` regular version and `<<-` where all leading tabs are ignored.

Here are the current proposals so we can get this conversation started.

## Proposal 1
```hlb
fs default() {
	image "alpine"
	run <<-END
	apk add -U foo \
		bar \
		baz
	END <<-END
	apk add -U foo \
		bar \
		baz
	END with option {
		env "key" "value"
	}
}
```

## Proposal 2
```hlb
fs default() {
	image "alpine"
	run <<-ARG1 <<-ARG2 with option { env "key" "value"; }
	apk add -U foo \
		bar \
		baz
	ARG1
	apk add -U foo \
		bar \
		baz
	ARG2
}
```

PR #107, opened 2020-04-18
[#101] implement heredoc support

coryb:
Adds support for:
```
value <<EOM
  this
    is 
  literal
EOM

# undent with <<-
value <<-EOM
  this
    is
  undented
EOM

# fold to single line with <<~
value <<~EOM
  this 
  is
  folded
EOM
```


parser/cst.go
```diff
@@ -24,20 +27,74 @@ var (
...
+	hereIdent lexer.Token
+}
+
+func (s *hereDocLexer) Next() (lexer.Token, error) {
+	t, err := s.lex.Next()
+	switch t.Type {
+	case s.sym["Whitespace"]:
+		if !s.keepSpace {
+			// not preserving spsace, so move on to next token
+			return s.Next()
```
hinshun, line 79: This is pretty elegant!

parser/unparse.go
```diff
@@ -348,8 +365,15 @@ func (s *BlockStmt) String() string {
 			}
 		}
 
+		insideHeredoc := false
 		for i := 1; i < len(stmts); i++ {
-			if len(stmts[i]) > 0 {
+			if stmts[i] == "<HLB-RAW-HEREDOC-START>" || stmts[i] == "<HLB-RAW-HEREDOC-END>" {
```
hinshun, line 370: Seems like we only need one marker? Can we move this to a constant at the top of this package. Maybe `<HLB-HEREDOC-MARKER>`.

coryb: fixed.

hinshun:
:rocket:

PR #110, opened 2020-04-19
add template processing builtin

coryb:
adding a basic templating function.  can be used like:
```
string secretPath() {
  template <<-EOM
    {{- if . -}}
      /path/to/ci/secret
    {{- else -}}
      /path/to/user/secret
    {{- end -}}
  EOM string {
    localEnv "CI"
  }
}
```

The part I don't like about this is the argument data into the template.  The PR currently passes the variadic arguments as a `[]string` slice when there is more than one argument to template, otherwise it will pass just a single `string` argument if only 1 variadic arg.  I am not sure there is a more logical approach.  If we wanted to pass named arguments to allow for `{{ .name }}` we would have to have some syntax that supports maps, which we dont have yet.


language/builtin.hlb
```diff
@@ -500,6 +500,16 @@ string format(string formatString, variadic string values)
 # @return the environment variable's value
 string localEnv(string key)
 
+# Process text as a Go text template with the arguments
+# assigned to the template data as "."
+# For template syntax documentation see:
+#   https://golang.org/pkg/text/template/
+#
+# @param the text of the template
```
hinshun, line 508: @param needs to be followed by `text`

coryb:
> ```
> string secretPath() {
>   template <<-EOM
>     {{- if .CI -}}
>       /path/to/ci/secret
>     {{- else -}}
>       /path/to/user/secret
>     {{- end -}}
>   EOM with option {
>         stringField "CI" string { localEnv "CI"; }
>   }
> }
> ```

I thought about it a while, I think this makes the most sense.  It will be less cumbersome when we fix #100 and I think it provides flexibility going forward.

I will update the PR for `option::template`

language/builtin.hlb
coryb: fixed

coryb:
@hinshun updated PR with option::template

PR #112, opened 2020-04-20
add localOs, localArch, and localCwd builtins

coryb:
I want to have a `hlb run` that automatically cross compiles for my current platform (darwin) but I need to pass in my local OS and Arch to the build graph.

Also I think there are some use-cases where `cwd` is relevent for building/exporting files back to the local workspace. `gdb` for example will only read source files if the strings embedded in the binary headers match the local system paths, so debugging a binary built in `/src` will not be easy if the sources are actually in `$HOME/project/src`.  With `localCwd` it will allow for patterns like:
```
fs build() {
  run "make" with option {
    dir localCwd
    mount fs { local "."; } localCwd
  }
}
```

### DIFF
--- a/builtin/builtin.go
+++ b/builtin/builtin.go
@@ -434,10 +434,19 @@ var (
 							parser.NewField(parser.Str, "values", true),
 						},
 					},
+					"localArch": FuncLookup{
+						Params: []*parser.Field{},
+					},
+					"localCwd": FuncLookup{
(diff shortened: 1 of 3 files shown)


hinshun:
My worry is that `runtime.GOOS` and `runtime.GOARCH` will be only useful for go development. Let me see how BuildKit Dockerfiles are handling this.

PR #115, opened 2020-04-21
Coerce fs func lit into group func lit

hinshun:
- Coerce fs func lit into group func lit
- Add `treeprint` branches for solve options so that they are testable
- Add codegen test

```
group default() {
	parallel group {
		image "alpine"
	} fs {
		scratch
		mkfile "foo" 0o644 "hello world"
		download "."
	}
}
```

### DIFF
--- a/codegen/codegen_test.go
+++ b/codegen/codegen_test.go
@@ -140,8 +140,8 @@ func TestCodeGen(t *testing.T) {
 		[]string{"default"},
 		`
 		group default() {
-			image "alpine"
-			image "busybox"
+			parallel fs { image "alpine"; }
+			parallel fs { image "busybox"; }
 		}
 		`,
 		func(t *testing.T, cg *CodeGen) solver.Request {
@@ -155,9 +155,9 @@ func TestCodeGen(t *testing.T) {
 		[]string{"default"},
 		`
 		group default() {
-			parallel group {
+			parallel fs {
 				image "alpine"
-			} group {
+			} fs {
 				image "busybox"
 			}
 		}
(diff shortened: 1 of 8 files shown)


codegen/codegen_test.go
```diff
@@ -210,6 +211,29 @@ func TestCodeGen(t *testing.T) {
...
 			)
 		},
+	}, {
+		"parallel coercing fs to group",
+		[]string{"default"},
+		`
+		group default() {
+			parallel group {
+				image "alpine"
+			} fs {
```
coryb, line 221: I think this should be a syntax error, I don't think `fs` builtins should be allowed within a group literal.  If they are allowed I would expect this to work:
```
parallel group {
  image "alpine"
  run "touch foo"
}
```
which won't work due to `run` starting with `scratch` state.  I think the only thing allowed in a group literal is `parallel`, `fs` literals and maybe more `group` literals.

This should work though:
```
parallel fs {
  image "alpine"
} fs {
  ...
}
```

coryb: We can update this in a subsequent PR

hinshun: I can do this in the same PR. Just a moment.

hinshun: To be clear, I will be adding:
1. Return checker error when `fs` builtin / functions are called within a group block.

And I will not be adding
1. Allowing `fs` literals to be defined in a `group` block. This one will be tougher to contribute.

coryb: yeah, that makes sense.

PR #121, opened 2020-04-24
use consistent local unique id for consistent local sha

coryb:


codegen/codegen_test.go
```diff
@@ -76,7 +76,7 @@ func TestCodeGen(t *testing.T) {
 		}
 		`,
 		func(t *testing.T, cg *CodeGen) solver.Request {
-			id, err := cg.LocalID(".")
+			id, err := cg.LocalID(context.Background(), ".")
```
hinshun, line 79: Can we deduplicate this by making a `ctx := context.Background()` at the top?

coryb: fixed

PR #123, opened 2020-04-24
Fix merging user defined option::copy with func lit

hinshun:
- The with options from `llb` upstream only takes `&llb.CopyInfo{}`, this means that merging additional user-defined options overwrites with zero values. I.e. `&llb.CopyInfo{CreateDestPath: true}` + `&llb.CopyInfo{AllowWildcard: true}` = `&{llb.CopyInfo{AllowWildcard: true}`.
- Also adds `option::copy chmod` because I discovered it was also available.
- We should probably get this upstream, I made a issue in BuildKit tracking this here: https://github.com/moby/buildkit/issues/1461


language/builtin.hlb
```diff
@@ -445,6 +445,12 @@ option::copy allowEmptyWildcard()
 # @return an option to change the owner of the copy path.
 option::copy chown(string owner)
 
+# Modifies the permissions of the copied files.
+#
+# @param filemode the new permissions of the file.
```
coryb, line 450: `@param owner`

hinshun: Nice catch! Though it should be `@param filemode` and `int filemode`!

PR #124, opened 2020-04-26
Allow option funcs to be used from imported

hinshun:
- Previously didn't allow `run "echo foo" with myimport.foo`
- We need some way of running codegen tests with imports, but going to take more work since we need a sandboxed buildkitd or some small mock framework for imports.


cmd/hlb/command/app.go
```diff
@@ -18,7 +18,7 @@ func App() *cli.App {
 
 	defaultAddress := os.Getenv("BUILDKIT_HOST")
 	if defaultAddress == "" {
-		defaultAddress = appdefaults.Address
+		defaultAddress = appdefaults.
```
coryb, line 21: looks like an accidental edit?

hinshun: My bad.

PR #125, opened 2020-04-27
Test codegen basic cases

### LINKED ISSUE #88: Unit test bundle for codegen basic cases
Tests should go in `codegen/codegen_test.go`. The input program should be as minimal as possible. There may be multiple test cases per bullet point:

- [x] scratch
- [x] http
- [x] git
- [x] env
- [x] dir
- [x] user
- [x] mkdir
- [x] mkfile
- [x] rm
- [x] copy

antonio-osorio:
Resolves #88

### DIFF
--- a/codegen/codegen_test.go
+++ b/codegen/codegen_test.go
@@ -41,6 +41,7 @@ func cleanup(value string) string {
 }
 
 func TestCodeGen(t *testing.T) {
+	t.Parallel()
 	ctx := context.Background()
 	for _, tc := range []testCase{{
 		"image",
@@ -397,8 +398,6 @@ func TestCodeGen(t *testing.T) {
 	}} {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
-
 			cg, err := New()
 			require.NoError(t, err)
 
--- a/codegen/codegen_test.go
+++ b/codegen/codegen_test.go
@@ -123,6 +123,18 @@ func TestCodeGen(t *testing.T) {
 				llb.KeepGitDir()))
 		},
 	}, {
+		"basic env",
+		[]string{"default"},
+		`
+		fs default() {
+			scratch
+			env "TEST_VAR" "test value"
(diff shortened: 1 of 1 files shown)


codegen/codegen_test.go
```diff
@@ -53,6 +54,86 @@ func TestCodeGen(t *testing.T) {
...
+		"basic env",
+		[]string{"default"},
+		`
+		fs default() {
+			scratch
+			env "TEST_VAR" "test value"
+		}
+		`,
+		func(t *testing.T, cg *CodeGen) solver.Request {
+			return Expect(t, llb.Scratch().AddEnv("TEST_VAR", "test value"))
```
antonio-osorio, line 135: @hinshun Heads up that this is actually a no-op. The environment is taken into consideration when displaying the Tree.

hinshun: Yeah, that's fine. I'm okay with this.

hinshun:
@antonio-osorio Are you still working on this? Are you planning to split off the rest of the items of #88 in a different PR?

antonio-osorio:
@hinshun I have been using the work-in-progress label for the issues that I would like some feedback on (hence creating the PR) but not yet completed.

This is ready for merge now. But please note that the `env`, `dir` and `user` tests are no ops since they are not reflected in the Tree representation.

hinshun:
@antonio-osorio If you have an empty `run ""` statement after, it'll inherit all those properties and be reflected in the Tree representation.

I think we can use the "Draft PR" feature, it seems to be what you're looking for, and I can still review them.

antonio-osorio:
> I think we can use the "Draft PR" feature, it seems to be what you're looking for, and I can still review them.

Had not seen it before. I have converted the PR to a draft

antonio-osorio:
> @antonio-osorio If you have an empty `run ""` statement after, it'll inherit all those properties and be reflected in the Tree representation.

`run ""` is not a valid statement, added an innocuous command.

PR #133, opened 2020-04-30
fix template for reference.md

coryb:
Also adding a step to PR build to verify generated files are up-to-date

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ env:
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: buildkitd
     steps:
     - name: Set up Go 1.13.10
       uses: actions/setup-go@v1
@@ -23,13 +23,6 @@ jobs:
 
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
-
-    - name: Restore Cache
-      uses: actions/cache@v1
-      id: cache
-      with:
-        path: ~/go/pkg/mod
-        key: ${{ runner.os }}-${{ hashFiles('**/go.sum') }}
 
     - name: Get dependencies
       run: go get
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -30,17 +32,17 @@ jobs:
         path: ~/go/pkg/mod
         key: ${{ runner.os }}-${{ hashFiles('**/go.sum') }}
 
-    - name: Get dependencies
-      run: go get
+    - name: Vendor
+      run: go mod vendor
```
hinshun, line 36: Curious why we chose to go the vendoring route here?

coryb: It has been faster for me to vendor and import vendored directory to buildkit rather than letting each container go get and rebuild from cache.  The go.hlb build will still force go-get, but when we move to `openllb/go.hlb` I was planning to move to vendor for those builds also.  The goal is to only resolve the deps once for each PR build.

hinshun: Using a mount for caching doesn't have the same effect?

coryb: I have removed the vendor stuff, we can work this out later.

.github/workflows/test.yml
```diff
@@ -8,11 +8,12 @@ on:
 
 env:
   GO111MODULE: on
+  BUILDKIT_HOST: docker-container://buildkitd
 
 jobs:
   build:
     name: Build
-    runs-on: ubuntu-latest
+    runs-on: buildkitd
```
hinshun, line 16: Wait what is this magic?

coryb: custom builder on digital ocean for better caching builds, here is the script we used to genreate the builder:
https://github.com/openllb/hlb/pull/133/commits/90b70c2b07e95d01c3699ddb8d71b12cbeea81ee

PR #137, opened 2020-05-02
add localRun command with options

coryb:


coryb:
> I think this PR looks good, but I've been thinking instead of the magic shell-style quoting deciding whether it runs with shell or not, we should probably always run with shell, and add a `option::run exec` or `option::run noShell` option to deliberately choose this behaviour. WDYT?

Hmm, so the rules would be:
1) if multiple variadic arguments then we directly execute
ie: `run "echo" "$HOME"` would run without a /bin/sh wrapper because we explicitly separate the arguments and run `["/bin/echo", "$HOME"]`

2) if single argument always run with shell (barring noShell option)
ie `run "echo $HOME"` and `run "true"` would always run as `["/bin/sh",  "-c", "echo $HOME"]` and `["/bin/sh", "-c", "true"]`

Maybe we can have `shlex` option to force lexing like we do now? So:
`run "sleep 10" with shlex` would run `["/bin/sleep", "10"]` rather than `["/bin/sh", "-c", "sleep 10"]` and `run "true" with shlex` would run `["/bin/true"]`

hinshun:
Yeah that works for me.

coryb:
@hinshun updated PR to include `shlex` option for both `run` and `localRun`

PR #141, opened 2020-05-08
Allow local to specify a single file instead of a directory

hinshun:
- Currently, if you specify a single file like `local "manifest.json"` then it fails with:
```sh
ERROR   failed to solve: rpc error: code = Unknown desc = error from sender: manifest.json is not a directory
```
- Instead `local "manifest.json"` should be the equivalent of: `local "." with option { includePatterns "manifest.json"; }`, with the caveat that if the `path` is for a file then include and exclude patterns are ignored.

### DIFF
--- a/codegen/chain.go
+++ b/codegen/chain.go
@@ -232,10 +232,24 @@ func (cg *CodeGen) EmitFilesystemBuiltinChainStmt(ctx context.Context, scope *pa
 			return fc, err
 		}
 
+		fi, err := os.Stat(path)
+		if err != nil {
+			return fc, err
+		}
+
 		var opts []llb.LocalOption
 		for _, iopt := range iopts {
 			opt := iopt.(llb.LocalOption)
 			opts = append(opts, opt)
+		}
+
+		if !fi.IsDir() {
+			filename := filepath.Base(path)
+			path = filepath.Dir(path)
+
+			// When path is a filename instead of a directory, include and exclude
+			// patterns should be ignored.
+			opts = append(opts, llb.IncludePatterns([]string{filename}), llb.ExcludePatterns([]string{}))
 		}
 
 		id, err := cg.LocalID(ctx, path, opts...)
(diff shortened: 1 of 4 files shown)


codegen/chain.go
```diff
@@ -232,12 +232,26 @@ func (cg *CodeGen) EmitFilesystemBuiltinChainStmt(ctx context.Context, scope *pa
...
 			opts = append(opts, opt)
 		}
 
+		if !fi.IsDir() {
+			filename := filepath.Base(path)
+			path = filepath.Dir(path)
+
+			// When path is a filename instead of a directory, include and exclude
+			// patterns should be ignored.
+			opts = append(opts, llb.IncludePatterns([]string{filename}), llb.ExcludePatterns([]string{}))
```
coryb, line 252: so it is the last Include/ExcludePatterns option in the list that wins?  They are not merged?

codegen/codegen_test.go
```diff
@@ -343,10 +354,21 @@ func TestCodeGen(t *testing.T) {
...
-
-			return Expect(t, llb.Local(id, llb.SessionID(cg.SessionID()), llb.SharedKeyHint(".")))
+			return Expect(t, cg.Local(t, "."))
+		},
+	}, {
+		"local file",
+		[]string{"default"},
+		`
+		fs default() {
+			local "codegen_test.go"
```
coryb, line 364: we probably want to add a explicit test to verify that include/exclude patterns are (and continue to be) ignored:
```
fs default() {
  local "codegen_test.go" with option {
    includePatterns "ignored"
    excludePatterns "ignored"
  }
}
```

codegen/chain.go
hinshun: Yeah, under the hood BuildKit just sets the array rather than appending it.

https://github.com/moby/buildkit/blob/1e4c22119456d26e7818473290962bc4503c8256/client/llb/source.go#L312-L319

codegen/codegen_test.go
hinshun: https://github.com/openllb/hlb/pull/141/commits/b8b8d5f7b4419bca9d27a7b3c3ec8c6656db2902

PR #142, opened 2020-05-09
Fix remote imports that have frontend codegen

hinshun:
- Currently I can't import a frontend like so:
```hlb
import hlbgen from fs {
    frontend "hinshun/hlbgen"
}
```
- The remote resolver should pass the client and multiwriter it already has.


hinshun:
It's a small change and I want to create a PR internally, so going to merge this in early.

PR #147, opened 2020-05-17
add `always` builtin for `run` to force command to always run (but allow cached mounts)

coryb:
The implementation is a bit of a hack.  There does not seem to be a way via llb to make an exec call force a cache-miss other than `llb.IgnoreCache`, but that will also prevent any mounts from also using cache. 

So this PR adds a random env var so that the vertex will always hash to a new digest while still allowing mount cache to be used.


coryb:
Created issue to remove hack when possible.

PR #154, opened 2020-05-21
Add source mapping to codegen'd LLB

hinshun:
- Source maps runtime LLB errors back to HLB (incl. a stacktrace)
```sh
 => ERROR copy /foo /                                                                                                                                                 0.0s
------
 > copy /foo /:
------
foo.hlb:2
--------------------
   1 |     fs func1() {
   2 | >>> 	func2
   3 |     }
   4 |
--------------------
foo.hlb:6
--------------------
   4 |
   5 |     fs func2() {
   6 | >>> 	func3
   7 |     }
   8 |
--------------------
foo.hlb:11
--------------------
   9 |     fs func3(){
  10 |     	scratch
  11 | >>> 	copy fs {
  12 |     		image "alpine"
  13 |     	} "/foo" "/"
--------------------
failed to solve: rpc error: code = Unknown desc = failed to compute cache key: "/foo" not found: not found
```

## TODO
- [x] Waiting for BuildKit PR to merge: https://github.com/moby/buildkit/pull/1494
- [x] Need to fix source mapping for relative and remote imports 
- [x] Waiting for side effects PR to merge: https://github.com/openllb/hlb/pull/164


cmd/hlb/command/run.go
```diff
@@ -205,7 +206,14 @@ func Run(ctx context.Context, cln *client.Client, rc io.ReadCloser, opts RunOpti
 		return solveReq.Solve(ctx, cln, p.MultiWriter())
 	})
 
-	return p.Wait()
+	err = p.Wait()
+	if err != nil {
+		for _, source := range errdefs.Sources(err) {
+			source.Print(os.Stderr)
```
coryb, line 212: We should probably add an `ErrOutput os.Writer` to the RunOptions so we can use that here and plumb it through from newt.

PR #158, opened 2020-05-22
Fix parsing error with inputs not ending with newline

### LINKED ISSUE #135: comment at end of file causes parse error
input:
```
# this is ok
fs default() {
	scratch
}
# this is not ok
```

error:
```
$ ./hlb run ./bug.hlb
./bug.hlb:5:1: unable to parse
```

I have run into this a few times as I comment out the last function in a file while debugging and then suddenly get parse errors.

hinshun:
Fixes #135 

- Wraps the `io.Reader` with a `NewlinedReader` that appends a newline before the natural EOF.
- Removes complexity dealing with zero-newline inputs. I didn't see much value in unparsing one-liners to retain one-liners.

### DIFF
--- a/checker/checker_test.go
+++ b/checker/checker_test.go
@@ -369,6 +369,10 @@ func TestChecker_Check(t *testing.T) {
 				Name: "localEnv",
 			},
 		},
+	}, {
+		"no error when input doesn't end with newline",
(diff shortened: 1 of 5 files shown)


coryb:
Looks like we need to fix the unparse to ignore the last trialing newline?

PR #164, opened 2020-06-12
Side effects and Binds

slushie:
This PR introduces Side Effects on Function Declarations, and exposes a BindClause that consumes those side effects. 

Still a Work in Progress, feedback appreciated!


parser/cst.go
```diff
@@ -795,17 +806,48 @@ type With struct {
...
-	As    *As    `parser:"@@"`
-	Ident *Ident `parser:"@@"`
-	Func  *FuncDecl
-	Call  *CallStmt
+// BindClause represents the entire "as ..." clause on a CallStmt, with either a
+// Default side effect or a list of Binds.
+type BindClause struct {
+	Pos     lexer.Position
+	As      *As       `parser:"@@"`
+	Default *Ident    `parser:"  @@"`
```
hinshun, line 814: Should be `( @@`

parser/cst.go
```diff
@@ -795,17 +806,48 @@ type With struct {
...
+	List    *BindList `parser:"| @@"`
```
hinshun, line 815: Should be `| @@ )`

parser/cst.go
```diff
@@ -795,17 +806,48 @@ type With struct {
...
+	default:
+		return shiftPosition(b.Pos, 1, 0)
+	}
+}
+
+// BindList is a parenthetical list of Binds.
+type BindList struct {
+	Pos        lexer.Position
+	OpenParen  *OpenParen  `parser:"@@"`
+	List       []*Bind     `parser:"( ( Newline )? @@ ( \",\" ( Newline )?  @@ )* ( \",\" Newline )? )?"`
```
hinshun, line 835: This looks correct.

parser/cst.go
hinshun: Also, instead of `Default` I prefer `Ident`

parser/cst.go
```diff
@@ -795,17 +806,48 @@ type With struct {
...
+
+func (b *BindClause) Position() lexer.Position { return b.Pos }
+func (b *BindClause) End() lexer.Position {
+	switch {
+	case b.Default != nil:
+		return b.Default.End()
+	case b.List != nil:
+		return b.List.End()
+	default:
+		return shiftPosition(b.Pos, 1, 0)
```
hinshun, line 827: Should be panic?

slushie: I actually just cargo-culted that line from elsewhere. Changed it to a panic though ;)

codegen/chain.go
```diff
@@ -169,6 +174,15 @@ func (cg *CodeGen) EmitFilesystemBuiltinChainStmt(ctx context.Context, scope *pa
 	}
 
 	switch expr.Name() {
+	case "state":
```
hinshun, line 177: Intentionally left here?

checker/builtin.go
```diff
@@ -0,0 +1,50 @@
...
+				obj = &parser.Object{
+					Kind:  parser.DeclKind,
+					Ident: ident,
+					Node: &BuiltinDecl{
+						Ident: ident,
+						Func:  make(map[parser.ObjType]*parser.FuncDecl),
+					},
+				}
+			}
+			node, ok := obj.Node.(*BuiltinDecl)
```
hinshun, line 37: s/node/decl

codegen/chain.go
```diff
@@ -637,12 +651,46 @@ func (cg *CodeGen) EmitFilesystemBuiltinChainStmt(ctx context.Context, scope *pa
...
+				solver.WithCallback(func(resp *client.SolveResponse) error {
+					digest = resp.ExporterResponse[keyContainerImageDigest]
+					return nil
+				}),
+			)
 			if err != nil {
 				return st, err
 			}
+
+			// XXX will this be de-duped if already solved by the binding?
```
hinshun, line 665: Yes, if it's not too complicated we should avoid duping but buildkit should dedupe for us.

codegen/codegen.go
```diff
@@ -305,14 +301,14 @@ func (cg *CodeGen) EmitBlock(ctx context.Context, scope *parser.Scope, typ parse
...
-		chain, err := cg.EmitChainStmt(ctx, scope, typ, call, ac, v)
+		chain, err := cg.EmitChainStmt(ctx, scope, typ, call, v)
 		if err != nil {
 			return v, err
 		}
 
 		var cerr error
 		v, cerr = chain(v)
-		if cerr == nil || cerr == ErrAliasReached {
+		if cerr == nil || errors.As(cerr, &ErrBindingCycle{}) {
```
hinshun, line 311: Nice, I do like the new `errors.As`.

codegen/errors.go
```diff
@@ -18,6 +18,10 @@ var (
 	// object for a scope. Lexical scope checking should occur in the checker so
 	// that's often where the bug shuold be fixed, but it may be in the codegen.
 	ErrUndefinedReference = errors.Errorf("undefined reference")
+
+	// ErrNoValue is returned when codegen is expecting a side effect value, but
+	// none is available. This is usually caused by bad external dependencies.
+	ErrNoValue = errors.Errorf("no value")
```
hinshun, line 24: This looks unused?

solver/solve.go
```diff
@@ -63,7 +63,7 @@ func WithDownloadOCITarball() SolveOption {
 	}
 }
 
-func WithCallback(fn func() error) SolveOption {
+func WithCallback(fn func(resp *client.SolveResponse) error) SolveOption {
```
hinshun, line 66: We should probably make this into a type. Maybe `SolveCallback`? Up to you.

codegen/chain.go
slushie: :facepalm:

hinshun:
@slushie Here's a summary of things I've changed:
- Solve binding immediately instead of waiting till first use (this means that if the binding is unused but the parent function is called, its still solved. if the parent function is not called, then no solves are done)
- Solve binding once for all `EmitBinding`
- Return `ErrBindNoTarget` if you bind like `dockerPush "some/ref" as`
- If decls are duplicated with builtins, show where its duplicated in user code rather than `<builtin>:0:0`
- Fix syntax highlighting in langserver
- Fix some edge case checker/codegen issues and tests

I've tested this against more complicated HLBs and I think this is ready to go! Thank you for this amazing feature! :tada:

PR #178, opened 2020-08-15
Provide image metadata via "manifest" builtin

slushie:
This PR adds the `manifest` builtin, which returns a string containing the JSON-encoded image config for a given image ref. It also provides the image's digest as a side effect.

It is defined as:

```
string manifest(string ref) binds (string digest)
```

The following example shows how to fetch and use this data:
```
string alpineManifest() {
    manifest "alpine" as alpineDigest
}

fs default() {
    scratch
    mkfile "manifest.json" 0o666 alpineManifest
    mkfile "image.properties" 0o666 string {
        format "imageID=%s\n" alpineDigest
    }
    download "."
}
```


codegen/chain.go
```diff
@@ -1106,6 +1114,67 @@ func (cg *CodeGen) EmitStringChainStmt(ctx context.Context, scope *parser.Scope,
...
+				return llb.State{}, err
+			}
+			if dgst == "" {
+				return llb.State{}, fmt.Errorf("no digest available for ref %q", ref)
+			}
+			return llb.State{}, nil
+		})
+		req, err := cg.outputRequest(ctx, st, Output{}, solver.WithCallback(
+			func(ctx context.Context, resp *client.SolveResponse) error {
+				return nil
```
hinshun, line 1152: What's the purpose of this callback?

slushie: No purpose, I'll remove it ;)

slushie:
Per our earlier conversation, should `manifest` use the containerd image resolver to return the manifest _list_? Or should this be renamed to `imageConfig` or something along those lines, to coincide with the returned value being the Docker image config and not a manifest at all?

hinshun:
> Per our earlier conversation, should `manifest` use the containerd image resolver to return the manifest _list_? Or should this be renamed to `imageConfig` or something along those lines, to coincide with the returned value being the Docker image config and not a manifest at all?

Would prefer to raw manifest (which is OCI) rather than the docker image config (which is not OCI)

slushie:
With the latest changes, `manifest` can be used to fetch the plain digest, the image index (aka manifest list) and the image config. 

```
string resolveManifest(string ref) {
    manifest ref with option {
        platform "linux" "amd64"
    } as (
        digest resolveDigest,
        index resolveIndex,
        config resolveConfig)
}

fs default() {
	scratch
	mkfile "manifest.json" 0o666 string {
	    resolveManifest "alpine:latest"
	}
	mkfile "config.json" 0o666 string {
	    resolveConfig "alpine:latest"
	}
	mkfile "index.json" 0o666 string {
	    resolveIndex "alpine:latest"
	}
	mkfile "image.properties" 0o666 string {
	    format "imageID=%s\n" string {
	        resolveDigest "alpine:latest"
	    }
	}
	download "build"
}
```

PR #180, opened 2020-08-22
Simplify codegen with value-oriented refactor

hinshun:
Rewrite of codegen, now much closer to the theory used in discussion. First class value / type coercion, and return registers.

- `CodeGen` greatly simplified down to ~300 lines.
- `CodeGen` is decoupled from implementation of builtins.
- `CodeGen` emitters no longer return any values other than `error`. All emitters that can modify the register has access to `ret`, the return register. No more `chainStart`.
- `CodeGen` is no longer stateful for the current compiling target, that means we can parallelize codegen trivially. No more `cg.reset()`.
- Arguments and options are evaluated before invoking the function (same as theory now). They are evaluated with a new register, and no longer write to the scope. No more `ExprKind`.
- Values and type coercion are now first class, no more `interface{}` or dealing with golang casting.
- `llb.State` is now wrapped by `codegen.Filesystem`, it also holds `image.Config`, `[]SolveOption` and the new `[]SessionOption` instead of `CodeGen`.
- Non control flow metadata (program counter, stacktrace, source mapping, session ID) moved to `context.Context`.
- Refactored `sockprovider` to `codegen/sockproxy`
- Refactored `checker.FormatPos` to `parser.FormatPos`
- Removed output flags from CLI targets

Reflect playground for `codegen.Value`: https://play.golang.org/p/c7n3fw-N5pt


codegen/builtin.go
```diff
@@ -0,0 +1,129 @@
+package codegen
+
+type Function interface{}
+
+var BuiltinFunctions = map[string]Function{
```
slushie, line 5: Should these files be in the `builtin` package?

codegen/builtin_fs.go
```diff
@@ -0,0 +1,499 @@
...
+
+type Scratch struct{}
+
+func (s Scratch) Call(ctx context.Context, ret Value, opts Options) error {
+	return ret.Set(llb.Scratch())
+}
+
+type Image struct{}
+
+func (i Image) Call(ctx context.Context, ret Value, opts Options, ref string) error {
```
slushie, line 20: Could the `Call()` func belong to a `Callable` interface? Perhaps taking `args []Value` instead of splatting the args into the function signature

codegen/builtin_fs.go
```diff
@@ -0,0 +1,499 @@
...
+
+	fs, err := ret.AsFilesystem()
+	if err != nil {
+		return err
+	}
+
+	run := fs.State.Run(runOpts...)
+	for path, mount := range mounts {
+		mnt := run.GetMount(path)
+		// FIXME: set bind
```
slushie, line 225: How can this set the binding value? Where are the function registers defined? Maybe we need to pass a `CallingContext` or whatever into `Call()` to capture the return value and bind registers?

codegen/builtin_fs.go
hinshun: I started off with a `Callable` interface, but I realized it was prone to error and code duplication where every function needed to validate the length of `args []Value` and then handle each type coercion.

```go
func (i Image) Call(ctx context.Context, ret Value, args []Value, opts Options) error {
    if len(args) != 1 {
        return errors.WithStack(ErrCodeGen{ProgramCounter(ctx), ErrNumArgs})
    }

    ref, err := args[0].AsString()
    if err != nil {
        return err
    }
```

When I refactored to splatting args into signature already typed, it reduced the complexity of all the builtin implementations, and I found a few `args[<index>]` and length checks were wrong.

In addition, I wanted to support things like `t time.Time` in the signature to ensure that all times in codegen are parsed with the same RFC format string.

codegen/builtin.go
hinshun: Yeah I'm debating on moving the `builtin` package as well, so currently in flux. But as it stands, yes I should move this file over.

codegen/builtin_fs.go
hinshun: The latest commit brings binding values in. I wanted to bring it closer to the concept that bind clauses are just an alias for a function but changes the target for the return value. In the current iteration, binds is basically used as a switch case to decide what value to write into `ret Register`.

I think caching the values returned from bindings should be generalized to having a `InstructionCache` that caches exprs in general.

PR #186, opened 2020-09-22
Allow building and importing Dockerfiles directly into HLB

hinshun:
This is only an experiment, the idea is the further bridge the gap between Dockerfiles and HLBs, allowing for users to incrementally migrate or leverage HLB without rewriting Dockerfiles.

## Running a stage in a Dockerfile with `hlb run`
```sh
❯ cat Dockerfile
FROM alpine AS foo

❯ hlb run -t foo Dockerfile
[+] Building 0.3s (2/2) FINISHED
 => [internal] load metadata for docker.io/library/alpine:latest                                                                                                      0.2s
 => CACHED [1/1] FROM docker.io/library/alpine@sha256:185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321                                                0.0s
 => => resolve docker.io/library/alpine@sha256:185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321                                                       0.0s
```

## Importing a dockerfile and referencing a stage as if it was an exported function of a HLB module.
```sh
❯ cat foo.hlb
import dockerfile from "./Dockerfile"

fs default() {
	dockerfile.foo
}

❯ hlb run foo.hlb
[+] Building 0.2s (2/2) FINISHED
 => [internal] load metadata for docker.io/library/alpine:latest                                                                                                      0.1s
 => CACHED [1/1] FROM docker.io/library/alpine@sha256:185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321                                                0.0s
 => => resolve docker.io/library/alpine@sha256:185518070891758909c9f839cf4ca393ee977ac378609f700f60a771a2dfe321                                                       0.0s
```

## Multi-stage docker build with multiple dockerfiles
```hlb
import dockerfile from "./Dockerfile"
import dockerfile2 from "./Dockerfile2"

fs default() {
	dockerfile.foo
	copy dockerfile2.bar "/foo" "/"
}
```

### DIFF
--- a/codegen/builtin_fs.go
+++ b/codegen/builtin_fs.go
@@ -18,6 +18,8 @@ import (
 	shellquote "github.com/kballard/go-shellquote"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/client/llb"
+	"github.com/moby/buildkit/exporter/containerimage/exptypes"
+	"github.com/moby/buildkit/frontend/dockerfile/dockerfile2llb"
 	gateway "github.com/moby/buildkit/frontend/gateway/client"
 	"github.com/moby/buildkit/session/filesync"
 	"github.com/moby/buildkit/solver/pb"
@@ -260,16 +262,14 @@ func (f Frontend) Call(ctx context.Context, cln *client.Client, ret Register, op
 				return
 			}
 
-			if ref == nil {
-				fs.State = llb.Scratch()
-			} else {
+			if ref != nil {
 				fs.State, err = ref.ToState()
 				if err != nil {
 					return
 				}
 			}
 
-			imageSpec, ok := res.Metadata[llbutil.KeyContainerImageConfig]
+			imageSpec, ok := res.Metadata[exptypes.ExporterImageConfigKey]
 			if ok {
 				err = json.Unmarshal(imageSpec, fs.Image)
 				if err != nil {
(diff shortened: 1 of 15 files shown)


hinshun:
WDYT @coryb?

parse.go
```diff
@@ -59,13 +60,17 @@ func Parse(r io.Reader, opts ...ParseOption) (*parser.Module, *parser.FileBuffer
...
-		name = "<stdin>"
+		name = "/dev/stdin.hlb"
 	}
 	r = &parser.NewlinedReader{Reader: r}
 
 	fb := parser.NewFileBuffer(name)
 	r = io.TeeReader(r, fb)
 
+	if filepath.Ext(name) != ".hlb" {
+		return parser.ParseDockerfile(r)
```
retrohacker, line 71: If I'm reading this correctly, it treats every file that doesn't have a `.hlb` extension as a dockerfile. Should we be more explicit and limit it to `name == "Dockerfile" || filepath.Ext(name) == ".dockerfile"`?

hinshun: Unfortunately, Dockerfiles do not have any conventions for naming.

hinshun: Introducing an extension may now require users to rename their Dockerfiles which may break their pipelines built on top.
I'll lean towards maybe a scheme within the HLB import, but I'm still on the fence.

Examples:
```sh
hlb run -t foo dockerfile://TestDockerfile
```

```hlb
import dockerfile from "dockerfile://TestDockerfile"
```

hinshun: It does lose some of the magic like `docker build` that people are used to though.

slushie: I was thinking about this yesterday and I realized what we need here are _options_ for `import`. For example,

```
import f1 from "./Dockerfile" with dockerfileSyntax
# or
import f2 from "./Dockerfile.foo" with option {
  syntax "dockerfile"
}
```

The first option is short and sweet, the second is explicit and verbose. But they're both parseable with no syntax changes, they don't encode semantics into the string so we can annotate the CST entries, and they're forward-compatible with hlbv2 or whatever other syntax we need in the future, including access to experimental Dockerfile syntax as eg, `with experimentalDockerfileSyntax`.

codegen/builtin_fs.go
```diff
@@ -281,9 +281,123 @@ func (f Frontend) Call(ctx context.Context, cln *client.Client, ret Register, op
...
+		if mw != nil {
+			pw = mw.WithPrefix("", false)
+		}
+
+		return solver.Build(ctx, cln, s, pw, func(ctx context.Context, c gateway.Client) (res *gateway.Result, err error) {
+			caps := c.BuildOpts().LLBCaps
+			st, img, err := dockerfile2llb.Dockerfile2LLB(ctx, dt, dockerfile2llb.ConvertOpt{
+				Target:       info.Target,
+				MetaResolver: c,
+				// BuildArgs:         filter(opts, buildArgPrefix),
```
slushie, line 339: I wish we could provide build args as arguments to the `fs` states exported by the Dockerfile. But because build args are unordered key/value pairs, I think we'd still need the same type of syntax that we use for frontend inputs, eg: 
```
dockerfile.someStage with option {
  arg "foo" "bar"
}
```

With a bit of hackery though, the Dockerfile stages could take `variadic string` args and they could be parsed as `key=value`, eg `dockerfile.someStage "foo=bar" "baz=bop"`

parse.go
hinshun: Hey I think that's a neat idea!
You could just do `import f1 from "./Dockerfile" with syntax("dockerfile")` too.

That said, this specific PR is using the `Dockerfile2LLB` as a library, so we won't be able to adjust the syntax other than through the Dockerfile's syntax directive for the gateway (e.g. `# syntax=docker/dockerfile:experimental`). Due to this, we may even need to display the Dockerfile syntax version alongside `hlb version`.

hinshun: The issue with an opaque syntax option is that we don't know if its a Dockerfile (or what version), or some other frontend. And we currently don't allow invoking an imported module as a function. For Dockerfiles, we can import their stages as exported identifiers, but otherwise there aren't any.

1. Do we allow invoking imports? (My opinion: no)
```hlb
import f1 from "./foobar" with syntax("foobar")
fs default() { f1 }
```
2. Do we invoke imports with `gateway.v0`? (Special case for Dockerfile to parse stages?)

codegen/builtin_fs.go
hinshun: Yeah I kept them commented because I thought we could provide a builtin `dockerBuild` that has options to expose these. This `func dockerfile` was meant to be consumed by both `DockerFileDecl` (for `hlb run` and `import`)  and `dockerBuild` builtin.

parse.go
slushie: I think for (1) you're right, it feels "weird" to directly invoke the import, which would otherwise be a module identifier. Because Dockerfile stages don't require names, it might be good to define some automatic naming convention eg, the first stage is synthesized as `f1._1`, and maybe the default stage (ie, last stage) is `f1.default` in addition to whatever user-defined names the stages might have.

As for (2) well I hadn't considered it, but that _would_ be a pretty badass feature. It should probably follow the same naming conventions as the Dockerfile parser, though. 
So given `import f2 from "./build.earthly" with frontend("earth.ly/foo")` maybe its used as `f2.default` for a single Ref, and `f2._1, f2._2, ...` for multiple returned Refs.

hinshun:
Out of date, but I have some ideas on how to bring this idea further.

PR #189, opened 2020-09-29
Fix heredoc string interpolation

hinshun:
Source:
```hlb
fs default() {
	run <<~repro
                $k
        repro
}
```

Before:
```
❯ hlb format foo.hlb
foo.hlb:3:17: no lexer rules in state "Heredoc" matched input text "$k\n        repro..."
exit status 1
```

After:
```sh
❯ hlb format foo.hlb
fs default() {
	run <<~repro
                $k
        repro
}
```

### DIFF
--- a/parser/cst.go
+++ b/parser/cst.go
@@ -52,7 +52,7 @@ var (
 			{"HeredocEnd", `\b\1\b`, stateful.Pop()},
 			{"Whitespace", `\s+`, nil},
 			{"Interpolated", `\${`, stateful.Push("Interpolated")},
-			{"Text", `[^\s$]+`, nil},
+			{"Text", `\$|[^\s$]+`, nil},
 		},
 		"RawHeredoc": {
 			{"RawHeredocEnd", `\b\1\b`, stateful.Pop()},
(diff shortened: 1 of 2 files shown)


coryb:
Are we going to generate "unknown variable $k" in another PR?  Maybe add an error like:
foo.hlb:3:17: Unknown variable $k, did you mean to escape the Heredoc like "<<~`repro`"

hinshun:
@coryb As a lint warning? Cause `$k` is just a literal string, only `${k}` is interpolated.

coryb:
Ahh, I see, I was confusing the nature of the issue.  n/m

parser/cst.go
```diff
@@ -52,7 +52,7 @@ var (
 			{"HeredocEnd", `\b\1\b`, stateful.Pop()},
 			{"Whitespace", `\s+`, nil},
 			{"Interpolated", `\${`, stateful.Push("Interpolated")},
-			{"Text", `[^\s$]+`, nil},
+			{"Text", `\$|[^\s$]+`, nil},
```
retrohacker, line 55: Why `\$|[^\s$]` and not `[^\s]`?

retrohacker (approved):

parser/cst.go
hinshun: Because it will then consume the `$` needed for `${`.

PR #209, opened 2020-11-18
Allow pipelines to run any non-option functions

hinshun:
This will allow non-fs functions to be run as a pipeline (e.g. `string localRun` that has side effects).

```
pipeline default() {
    stage localRun("echo msg > foo")
    stage local("foo")
}
```

Another example, where a tarball is made locally before running a build that depends on the tarball made.
```
pipeline default() {
    stage localRun("./make-tarball.sh")
    stage buildDependingOnTarball
}
```


hinshun:
@coryb WDYT about this? In future, I want to be able to `hlb run -t <string-target>` as well, so this seems to fit in the theme of pipelines being to execute anything that is already a valid target.

coryb (approved):

PR #210, opened 2020-11-20
Add DAP support and decouple debugger from TUI

### LINKED ISSUE #11: Split off `codegen.Debugger` into its own package
Currently, the debugger sits inside the `codegen` package. We should refactor it out to a `debug` package.

hinshun:
Fixes #11 

- Adds DAP support over stdio when running with `hlb run --dap`. See: https://microsoft.github.io/debug-adapter-protocol/
- Decouples debugger from "debug frontends" like the TUI debugger. Now runs in a separate goroutine and has async coordination. Codegen debugger is written to be thread-safe

vscode


hinshun:
Superseded by #286 and #288

Issue #211, opened 2020-11-22
Proposing the composite type: arrays


hinshun:
Currently, the types supported are `string`, `int`, `bool`, `fs`, `option`, and `pipeline`.

This proposal introduces composite types, allowing for `[]string`, `[]int`, `[]bool`, `[]fs`, `[]option`, and `[]pipeline`, as well as `for loops`.

## Proposal
1. Statements in array blocks append to return register instead of modify.
2. Arrays are splatted automatically.

For arrays to work well in practice, we want to ensure we have data flow analysis to return compile errors for `dead code`, so that will be a prerequisite.
```hlb
string regions() {
	"us-east-1"
	^^^^^^^^^^^
	dead code
	"us-west-2"
}
```

Here is a function returning an array. Instead of overriding the value of  the return register, they are appended similar to declaring the elements of an array. (1. Statements in array blocks append to return register instead of modify)
```hlb
[]string regions() {
	"us-east-1"
	"us-west-2"
}
```

You can invoke functions that also return the same type array and they are splatted automatically. (2. Arrays are splatted automatically)
```hlb
[]string westRegions() {
	"us-west-1"
	"us-west-2"
}

[]string allRegions() {
	"us-east-1"
	westRegions
	^^^^^^^^^^^
	splatted
}
```

This is useful when combined with `for loop`, here's the proposed structure.
```hlb
for (<type> <ident> in <expr>) {
    # statements affect register as usual
}
```

For example, when you need to publish an image to multiple registries:
```hlb
fs publish() {
	build
	for (string region in regions) {
		dockerPush "registry-${region}.com/my-image"
	}
}
```

Since the `for loop` expects an expression, you can also provide a block literal as an expression.
```hlb
fs publish() {
	build
	for (string region in []string {
        	"us-east-1"
		westRegions
	}) {
		dockerPush "registry-${region}.com/my-image"
	}
}
```

Note you can still do this inline, but follows regular block syntax (i.e. separate statements by `;`).
```hlb
fs publish() {
	build
	for (string region in []string { "us-east-1"; "us-west-2" }) {
	                                            ^
	                                            odd compared to other languages
	                                            but consistent within HLB
		dockerPush "registry-${region}.com/my-image"
	}
}
```

Elements or slices of an array can be accessed through array indexing. For example, if we need to bind the digest of one of the registries we pushed to.
```hlb
fs registryPush(fs ref) {
	for (string region in regions[1:]) {
[...]

PR #212, opened 2020-11-25
add template funcs

coryb:
This adds the sprig template funcs for general usage
Also I have added several docker image parsing template funcs which I have needed for various docker image publishing routines.


codegen/builtin_string.go
```diff
@@ -53,6 +55,64 @@ func (t Template) Call(ctx context.Context, cln *client.Client, ret Register, op
 	return ret.Set(buf.String())
 }
 
+func templateFuncs() template.FuncMap {
+	return map[string]interface{}{
+		// dockerDomain returns the domain/host information for the provided
+		// docker image
+		"dockerDomain": func(in string) string {
```
slushie, line 62: Does it have to be `dockerFoo`? Could this be `imageRefDomain` or something a bit more OCI-friendly? 😉

hinshun: Was chatting on slack, but I was suggesting since we had interpolation support in heredocs, I prefer if we use regular string functions rather than template functions.

Along with the [array support proposal](https://github.com/openllb/hlb/issues/211), you could do something like:
```hlb
[]string dockerRegistries() {
	"titusregistry.us-east-1.netflix.net"
	"titusregistry.us-west-2.netflix.net"
}

[]string pushTargets(string ref) {
	for (string registry in dockerRegistries) {
		"${registry}/${dockerPath(ref)}:${dockerTag(ref)}
	}
}

fs titusPush(string ref) {
	for (string target in pushTargets(ref)) {
		dockerPush target
	}
}
```

PR #214, opened 2020-12-11
Cache image config resolves within a single hlb run

hinshun:
- Previously, we resolve for every `image` call, even if its a duplicate. Since resolve is slow, this may end up being the slowest part of the cached builds.

### DIFF
--- a/codegen/resolver.go
+++ b/codegen/resolver.go
@@ -1,0 +1,77 @@
+package codegen
+
+import (
+	"context"
+	"sync"
+
+	"github.com/docker/buildx/util/progress"
+	"github.com/moby/buildkit/client"
+	"github.com/moby/buildkit/client/llb"
+	gateway "github.com/moby/buildkit/frontend/gateway/client"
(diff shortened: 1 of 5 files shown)


jamesmulcahy:
❤️

codegen/resolver.go
```diff
@@ -0,0 +1,78 @@
...
+type imageConfig struct {
+	dgst   digest.Digest
+	config []byte
+}
+
+func (r *cachedImageResolver) ResolveImageConfig(ctx context.Context, ref string, opt llb.ResolveImageConfigOpt) (dgst digest.Digest, config []byte, err error) {
+	r.mu.RLock()
+	cfg, ok := r.cache[ref]
+	if ok {
+		r.mu.RUnlock()
```
coryb, line 39: I would move the `r.mu.RUnlock()` before the `if ok` check so you can dedup and remove the one on line 42

coryb (approved):
One minor nit, otherwise looks good!

Issue #217, opened 2020-12-17
BuildKit should support errors.Cause over GRPC


hinshun:
Currently some errors like `ENOENT` gets wrapped and error checks like `os.IsNotExist` or `errors.Is(err, os.ErrNotExist)` fail because they are unable to unwrap or "cause" the error after it's been through the BuildKit grpc error wrappers/unwrappers.

PR #218, opened 2020-12-17
Fix multiple issues with resolving imports

hinshun:
- `--target run` says "arguments required" because its trying to execute the
builtin run. we shouldn't allow them to be viable targets
- `no such file or directory` has a race condition where stratum-agent was used
to lookup common.hlb when its supposed to be looking in stratum-hlb-service
- `no such file or directory` doesn't show a good error because buildkit
upstream has a bug, but i have a workaround

Follow up: https://github.com/openllb/hlb/issues/217

### DIFF
--- a/module/resolve.go
+++ b/module/resolve.go
@@ -389,6 +389,7 @@ func resolveGraph(ctx context.Context, info *resolveGraphInfo, res Resolved, mod
 
 	parser.Match(mod, parser.MatchOpts{},
 		func(id *parser.ImportDecl) {
+			res := res
 			g.Go(func() error {
 				var (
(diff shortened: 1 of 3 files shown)


module/resolve.go
```diff
@@ -389,6 +389,7 @@ func resolveGraph(ctx context.Context, info *resolveGraphInfo, res Resolved, mod
 
 	parser.Match(mod, parser.MatchOpts{},
 		func(id *parser.ImportDecl) {
+			res := res
```
coryb, line 392: interesting, I would have expected the linters to complain about this issue

coryb (approved):

PR #229, opened 2021-01-15
update identifier regex to allow underscores

slushie:
This is a small PR to update the syntax highlighter to allow `_` (underscores) within identifiers.


hinshun (approved):

hinshun:
@coryb Does the vscode plugin need to be updated?

PR #230, opened 2021-01-26
Unparse BuiltinDecl correctly

aaronlehmann:
This fixes two problems:

1) `(*SpanError).Pretty` would panic if there were no spans for the
file. This seemed to be an oversight, because the length of `spans` was
checked just below.

2) `Module`'s implementation of `Unparse` was used for `BuiltinDecl`.
This meant that unrecognized builtin errors would end up printing the
whole builtin module instead of just the name of the builtin.

### DIFF
--- a/codegen/codegen_test.go
+++ b/codegen/codegen_test.go
@@ -905,3 +905,57 @@ func TestCodeGen(t *testing.T) {
 		})
 	}
 }
+
+func TestCodegenError(t *testing.T) {
+	t.Parallel()
+
+	type errorTestCase struct {
+		name          string
+		targets       []string
+		input         string
+		expectedError string
+	}
+
+	for _, tc := range []errorTestCase{
+		{
+			"invalid builtin",
+			[]string{"default"},
+			`
+			fs default() {
+				bug includePatterns("*.go")
+			}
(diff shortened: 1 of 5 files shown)


codegen/codegen_test.go
```diff
@@ -905,3 +905,57 @@ func TestCodeGen(t *testing.T) {
...
+				local "." with pattern
+			}
+		`,
+			"<stdin>:3:9: unrecognized builtin `includePatterns`",
+		},
+	} {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := diagnostic.WithSources(context.Background(), builtin.Sources())
+			mod, err := parser.Parse(context.Background(), strings.NewReader(tc.input))
```
hinshun, line 937: This should also use `ctx` from above. Sources parsed by `parser.Parse` will added via the `ctx`.

hinshun (commented):
Only one thing, otherwise looks great!

codegen/codegen_test.go
aaronlehmann: Thanks! - fixed

hinshun (approved):

PR #231, opened 2021-02-05
Write to log console instead of directly to os.Stderr

hinshun:
- Previously it directly wrote to `os.Stderr`
- Should probably provide some kind of helper to noop wrap `io.Writer` -> `console.File` in a follow up


cmd/hlb/command/run.go
```diff
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/containerd/console"
```
coryb, line 10: It seems like we are importing this package simply for an interface.  I vote for just adding out own interface here:
```
type File interface {
	io.ReadWriteCloser
	Fd() uintptr
	Name() string
}
```

hinshun: https://github.com/openllb/hlb/pull/232

PR #232, opened 2021-02-05
Remove direct dependency on containerd/console

hinshun:
Follow up from: #231

### DIFF
--- a/solver/progress.go
+++ b/solver/progress.go
@@ -3,10 +3,10 @@ package solver
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"time"
 
-	"github.com/containerd/console"
 	"github.com/docker/buildx/util/progress"
 	"github.com/moby/buildkit/client"
 	"github.com/moby/buildkit/identity"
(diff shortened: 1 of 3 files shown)


solver/progress.go
```diff
@@ -15,10 +15,20 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+type Console interface {
```
aaronlehmann, line 18: Maybe it would be good to have a comment that this is intended to match the `File` interface from `github.com/containerd/console`, to avoid pulling in that dependency. If that interface ever changes, we'll have to keep this copy in sync, and it will be easier to understand what happened with the comment giving context.

aaronlehmann (approved):

aaronlehmann (approved):

solver/progress.go
coryb: Well, it is an interface that matches `*os.File`, just that `console` implemented that interface.  Not sure we care about console as much as matching `*os.File`?

coryb (approved):

solver/progress.go
aaronlehmann: It needs to match what `progress.NewPrinter` expects as an argument.

PR #234, opened 2021-03-17
Frontend errgroup was not waited as a solve option

hinshun:
- Missed as part of a refactor sometime ago


coryb (approved):
LGTM

hinshun:
Actually think I need a different fix, because the errgroup wait only happens on final solve which is fine for `download` or `dockerPush` but not `frontend`.

We need to use either async llb or just wait for the errgroup here.

hinshun:
Opted for waiting for errgroup directly because we handle not only `llb.State` but also the image config (outside of LLB but `codegen.Filesystem` does handle).

PR #237, opened 2021-04-10
Add v1 compatibility container-config

hinshun:
Some systems depend on the `container_config` from v1 docker manifests (not present in official OCI image specs). Adding a compat layer during `dockerPush` evaluation.

Specifically, it's distributions manifest payload when requesting with `application/vnd.docker.distribution.manifest.v1+json`, activating this code:
https://github.com/distribution/distribution/blob/6affafd1f030087d88f88841bf66a8abe2bf4d24/manifest/schema1/config_builder.go#L183-L190

Modifying last history v1 compatibility with payload from the raw manifest config. This PR allows `container_config` to be plumbed into registry's automatic v1 translation to history with v1 compat.

### DIFF
--- a/solver/solve.go
+++ b/solver/solve.go
@@ -26,8 +26,23 @@ type SolveInfo struct {
 	OutputLocalTarball    bool
 	OutputLocalOCITarball bool
 	Callbacks             []SolveCallback `json:"-"`
-	ImageSpec             *specs.Image
+	ImageSpec             *ImageSpec
 	Entitlements          []entitlements.Entitlement
+}
+
+// ImageSpec is HLB's wrapper for the OCI specs image, allowing for backward
+// compatible features with Docker.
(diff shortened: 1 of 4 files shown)


solver/solve.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 
 	"github.com/docker/buildx/util/progress"
+	"github.com/docker/docker/api/types/container"
```
aaronlehmann, line 8: No particular opinion either way, but if you feel it's worth avoiding the dependency on `docker/docker` (which is a fairly large repo IIRC), you could use something like:

```go
type ContainerConfig struct {
        Cmd      []string
        Labels   map[string]string
}
```

aaronlehmann (approved):

aaronlehmann (approved):

PR #238, opened 2021-04-14
skip fetching rootfs blobs for `manifest`

coryb:
it seems the `manifest` routine was fetching all the image blogs including the data blobs.  This PR filters out any blob with `rootfs` in the type, so it will fetch just the image metadata.  It seems to fix my issue with the function hanging.

### DIFF
--- a/pkg/imageutil/buffered_resolver.go
+++ b/pkg/imageutil/buffered_resolver.go
@@ -2,6 +2,7 @@ package imageutil
 
 import (
 	"context"
+	"strings"
 
 	"github.com/containerd/containerd/content"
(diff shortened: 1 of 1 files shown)


aaronlehmann:
Wow. That's surpring behavior.

LGTM as far as I can tell

coryb:
It is a little hard to follow, but I think we did it to ourselves by using `FetchHandler` with the goal to fetch json document for the metadata.  It appears this handler does not look at the content type, it just reads all the data associated and we just ignore the ones not for the json documents.  I suspect this was only tested with busybox or alpine :)

hinshun:
lgtm

PR #252, opened 2021-07-16
Avoid duplicated log output from parallel solves

aaronlehmann:
When a parallel solve involves overlapping vertices, HLB may print log
lines multiple times.

This works around the issue by adding our own `MultiWriter` implementation
which keeps track of which `Writer` returned by `WithPrefix` is responsible
for outputting logs for any particular vertex. Writes for the same
vertex will only be allowed through once.

Repro case:

    fs default() {
            output "hi"
            download "."
    }

    fs common(string msg) {
            image "docker-hub.netflix.net/library/busybox"
            run "for i in $(seq 1 10); do echo $i; sleep 1; done" with ignoreCache
            run "echo ${msg} > /output/${msg}" with option {
                    mount scratch "/output" as output
            }
    }

### DIFF
--- a/solver/multiwriter.go
+++ b/solver/multiwriter.go
@@ -1,0 +1,80 @@
+package solver
+
+import (
+	"strings"
+	"sync"
+
+	"github.com/docker/buildx/util/progress"
+	"github.com/moby/buildkit/client"
+	digest "github.com/opencontainers/go-digest"
+)
+
+// MultiWriter is similar to progress.MultiWriter, but deduplicates writes by
+// vertex.
+type MultiWriter struct {
+	w                  progress.Writer
+	allClaimedVertices map[digest.Digest]struct{}
+	claimedVerticesMu  sync.Mutex
+}
+
+func NewMultiWriter(pw progress.Writer) *MultiWriter {
+	if pw == nil {
+		return nil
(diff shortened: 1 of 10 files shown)


solver/multiwriter.go
```diff
@@ -0,0 +1,80 @@
...
+		for _, v := range filtered.Vertexes {
+			v.Name = addPrefix(p.pfx, v.Name)
+		}
+	}
+	p.mw.w.Write(filtered)
+}
+
+func addPrefix(pfx, name string) string {
+	if strings.HasPrefix(name, "[") {
+		return "[" + pfx + " " + name[1:]
```
slushie, line 77: Is this missing a trailing `]`?

coryb: No, this will take `addPrefix("foo", "[other]")` and generate `[foo other]`.  I wonder if `[foo] [other]` would be better here?  Not sure of the current usage we might have....

slushie: Aha, gotcha 👍

coryb (approved):
LGTM.  I suspect there will be some odd edge cases from this, but I can't think of any currently.  In any case, it is definately better than what we have now.

PR #253, opened 2021-07-23
Implement gateway exec for debugger (carry #192)

aaronlehmann:
This is based on #192, with a few changes:

- `InterruptibleReader` is replaced with InputSteerer, which steers input
  data into one of two pipes, depending whether we've exec'd a command.

- Allow` Exec` if stdin is not a terminal.

- Fixed a few crashes when debugger commands are given in an unexpected
  context.

- A little refactoring in the debugger and the run command, to enable
  external reuse.

- Minor cleanups (mostly simplifying printfs)


codegen/debug.go
```diff
@@ -504,3 +539,155 @@ func attr(dgst digest.Digest, op pb.Op) (string, string) {
...
+
+			res, err = c.Solve(ctx, gateway.SolveRequest{
+				Definition: def.ToPB(),
+			})
+			if err != nil {
+				return
+			}
+
+			ctr, err := c.NewContainer(ctx, gateway.NewContainerRequest{
+				Mounts: []gateway.Mount{{
```
hinshun, line 581: Out of scope for this PR, but wonder if we should respect mounts in the with-options if the cursor is on a `run` call.

Also should we respect existing state like env vars set on `llb.State`, current working directory, etc?

hinshun (approved):
Overall, looks good.

codegen/debug.go
aaronlehmann: I've added support for this in the latest commit. In addition to respecting those options when the cursor is on a `run` statement, I've added equivalent options for `breakpoint` (for example, to exec a shell with specific mounts available).

cmd/hlb/command/run.go
```diff
@@ -222,14 +233,14 @@ func Run(ctx context.Context, cln *client.Client, rc io.ReadCloser, info RunInfo
 		return err
 	}
 
-	if solveReq == nil || info.Debug || info.Tree {
```
aaronlehmann, line 225: I noticed that this caused hlb to exit early when debugging was turned on. This seems counter-intuitive to me, because I'm not sure why using the debugger means we want to skip the solve. I removed this for now, but please let me know if I'm missing something.

codegen/codegen.go
```diff
@@ -463,6 +452,12 @@ func (cg *CodeGen) EmitCallStmt(ctx context.Context, scope *parser.Scope, call *
 		}
 	}
 
+	// Yield before executing the next call statement.
+	err = cg.Debug(ctx, scope, call.Name, ret, opts)
```
aaronlehmann, line 456: I moved this down here from above so `opts` would be available. I'm not 100% sure that's safe, and would appreciate confirmation.

cmd/hlb/command/run.go
hinshun: In the beginning, it was useful to just debug the debugger without having to cancel the solve every time, so this is just leftover from that. Hopefully HLB will one day execute in a single gateway build function, in which case solving every time is the only option.

codegen/codegen.go
hinshun: I believe this is necessary but I want to point out that this changes the behavior from stepping in this order: call -> args, to args -> call. Since args need to be evaluated and passed by value, I think this change makes sense.

You should try out stepping through the debugger to see the difference.

aaronlehmann: Ah yes, I did notice that, but didn't realize it was a result of this change.

codegen/debug.go
```diff
@@ -504,3 +542,237 @@ func attr(dgst digest.Digest, op pb.Op) (string, string) {
...
+				}
+
+				ctrReq.Mounts = append(ctrReq.Mounts, gateway.Mount{
+					Dest:      mount.Target,
+					MountType: pb.MountType_BIND,
+					Ref:       res.Ref,
+					Readonly:  mount.IsReadonly(),
+				})
+			}
+			for _, secret := range secrets {
```
aaronlehmann, line 645: @coryb: Does this look like the right way to attach a secret for gateway exec?

language/builtin.hlb
```diff
@@ -218,6 +218,17 @@ option::run secret(string localPath, string mountPoint)
...
+# Sets a debugger breakpoint just before the "run" command.
+#
+# @param command optional command to invoke for the debugger "exec" command,
+# followed by optional arguments
+# If no value is given, it will execute /bin/sh.
+# If exactly one value is given, the specified command will be invoked.
+# If more than one value is given, values after the first one are treated as
+# arguments for the command.
+# @return an option to set a breakpoint
+option::run breakpoint(variadic string command)
```
coryb, line 230: this implies `breakpoint` without arguments will work,  but I think we wanted `breakpoint(string command, variadic string args)` so as least one argument is required.

parser/cst.go
```diff
@@ -463,7 +463,10 @@ func NewCallStmt(name string, args []*Expr, with *WithClause, binds *BindClause)
 	}
 }
 
-func (cs *CallStmt) Breakpoint() bool {
+func (cs *CallStmt) Breakpoint(returnType Kind) bool {
+	if returnType == Kind(fmt.Sprintf("%s::%s", Option, "run")) {
```
coryb, line 467: This is no longer necessary right?

builtin/builtin.go
```diff
@@ -154,6 +156,16 @@ var (
 		"option::manifest": map[string]parser.Callable{
 			"platform": codegen.Platform{},
 		},
+		"option::breakpoint": map[string]parser.Callable{
```
coryb, line 159: this is no longer necessary?

language/builtin.hlb
aaronlehmann: I was thinking it should be okay to specify `breakpoint` without args and default to `/bin/sh` for an exec. That would keep things consistent with the `breakpoint` statement. Now that we've decided we won't use the `run` command as an exec default, I don't think there's ambiguity in what gets run. But I'm fine with requiring at least one arg if you prefer.

parser/cst.go
aaronlehmann: It is necessary now that I added the run option. Without it, `breakpoint` inside `with option {` gets treated as a breakpoint, and the debugger will break on that option (rather than just the `run`).

builtin/builtin.go
aaronlehmann: So far I haven't removed the `option::breakpoint` values... it wasn't clear to me from the other discussion whether we wanted to prevent `breakpoint` from taking run options, or prevent it from taking options at all. If these are unwanted, I can remove them.

language/builtin.hlb
coryb: okay, works for me 👍

parser/cst.go
coryb: I understand now, I was confusing this with trying to use option::run with the breakpoint instruciton.

aaronlehmann: Also please let me know if you can think of a cleaner way to do this!

builtin/builtin.go
coryb: Okay, I don't see a reason to keep these really, I think they just add complexity as they are redundant with `option::run breakpoint`.

coryb (approved):
This is awesome!  So nice to be able to get into a container and run adhoc commands.

PR #257, opened 2021-07-28
Add `Sync` to Progress to synchronize progress while debugging

coryb:
This will now show the "plain" progress while debugging.  The goal is to ensure all progress output is flushed at various points, this is essential while in the debugger especially before we exec into a container.   There is no way to sync the progress writers we use upstream, but we can wait for completion, so this PR creates a `Sync` func in our progress writer which basically shuts down the current progress writer and waits for completion, then recreates the progress writer so that we can continue to collect progress later.

### DIFF
--- a/codegen/debug.go
+++ b/codegen/debug.go
@@ -66,6 +66,8 @@ func NewDebugger(c *client.Client, w io.Writer, inputSteerer *InputSteerer, prom
 		breakpoints       []*Breakpoint
 	)
 
+	var previousCommand string
+
 	return func(ctx context.Context, scope *parser.Scope, node parser.Node, ret Value, opts Option) error {
 		// Store a snapshot of the current debug step so we can backtrack.
 		historyIndex++
@@ -94,7 +96,7 @@ func NewDebugger(c *client.Client, w io.Writer, inputSteerer *InputSteerer, prom
 			}
 
 			if next != nil {
-				// If nment is not in the same function scope, skip over it.
+				// If next is not in the same function scope, skip over it.
 				if next != fun {
 					return nil
 				}
@@ -107,6 +109,10 @@ func NewDebugger(c *client.Client, w io.Writer, inputSteerer *InputSteerer, prom
 			}
 
 			for {
+				err := Progress(ctx).Sync()
+				if err != nil {
+					return err
+				}
 				fmt.Fprint(w, "(hlb) ")
 
 				command, err := promptReader.ReadString('\n')
@@ -116,9 +122,12 @@ func NewDebugger(c *client.Client, w io.Writer, inputSteerer *InputSteerer, prom
 
 				command = strings.Replace(command, "\n", "", -1)
 
-				if command == "" {
+				if command == "" && previousCommand == "" {
 					continue
+				} else if command == "" {
+					command = previousCommand
 				}
+				previousCommand = command
 
 				args, err := shellquote.Split(command)
 				if err != nil {
@@ -248,7 +257,7 @@ func NewDebugger(c *client.Client, w io.Writer, inputSteerer *InputSteerer, prom
 							return
 						}
 					}()
-				case "exit":
+				case "exit", "quit", "q":
 					return ErrDebugExit
 				case "funcs":
 					for _, obj := range s.scope.Defined() {
@@ -736,6 +745,11 @@ func ExecWithFS(ctx context.Context, cln *client.Client, fs Filesystem, r io.Rea
 			}
 			defer ctr.Release(ctx)
 
+			err = Progress(ctx).Sync()
+			if err != nil {
+				return
+			}
+
 			startReq := gateway.StartRequest{
 				Args:         args,
 				Cwd:          cwd,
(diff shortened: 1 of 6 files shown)


codegen/debug.go
```diff
@@ -119,9 +125,12 @@ func NewDebugger(c *client.Client, w io.Writer, unbufferedReader io.Reader) Debu
 
 				command = strings.Replace(command, "\n", "", -1)
 
-				if command == "" {
+				if command == "" && previousCommand == "" {
 					continue
+				} else if command == "" {
+					command = previousCommand
 				}
+				previousCommand = command
```
coryb, line 133: This allows us to just hit enter to repeat the last command, it is frustrating typing `next\n` over and over to step through the program.  Most debuggers allow you to type `next\n`, then subsequent `\n` will repeat that command.

aaronlehmann: Should we limit it to commands where this makes sense... for example I don't think we'd want to repeat `break` or `exec`?

solver/progress.go
```diff
@@ -111,44 +116,56 @@ func NewProgress(ctx context.Context, opts ...ProgressOption) (Progress, error)
...
-		done: done,
-	}, nil
+func (p *progressUI) Sync() error {
+	p.Release()
+	err := p.Wait()
+	if err != nil {
+		return err
+	}
+	wait, release := p.spp.reset()
+	p.g, p.ctx = errgroup.WithContext(p.origCtx)
```
aaronlehmann, line 152: Does the write to `p.g` need to be synchronized with a mutex, to avoid racing with `p.Write`?

codegen/debug.go
coryb: Not sure, I think "always execute previous command" is easier to reason about than special-casing a few commands.  I believe this is consistent behavior with gdb and other debuggers I have used, although I am not 100% positive on that.

solver/progress.go
```diff
@@ -111,44 +116,56 @@ func NewProgress(ctx context.Context, opts ...ProgressOption) (Progress, error)
...
+	p.done = make(chan struct{})
+	p.g.Go(func() error {
+		defer release()
+		<-p.done
+		return wait()
+	})
```
aaronlehmann, line 158: Curious why we need a goroutine that just sits around and waits for a call to `Release`. Couldn't we just have `p.Wait()` block on `p.done` and call `wait`?

coryb: Yeah, this was mostly old stuff that I just tweaked a bit.  I have refactored to get rid of this gratuitous use of the errGroup.   I also ended up removing `Go` from the progress interface.  That usage was also gratuitous and prevented me from calling. Sync before exec'ing into the error container.

solver/progress.go
coryb: Good call, added a mutex and lock between Sync and Write

codegen/debug.go
aaronlehmann: I remember hating this gdb behavior, but I won't object.

codegen/debug.go
```diff
@@ -248,7 +257,7 @@ func NewDebugger(c *client.Client, w io.Writer, inputSteerer *InputSteerer, prom
 							return
 						}
 					}()
-				case "exit":
+				case "exit", "quit", "q":
```
aaronlehmann, line 260: Nice, I was thinking about doing the same for `continue` (`c`)

solver/progress.go
```diff
@@ -230,52 +234,55 @@ func (p *progressUI) Release() {
...
-
-	done := make(chan struct{})
-	g.Go(func() error {
-		<-done
-		return nil
-	})
+func (p *progressUI) Wait() (err error) {
+	defer p.spp.cancel()
+	defer func() {
+		gerr := p.g.Wait()
```
aaronlehmann, line 240: Isn't this redundant with the Wait at the end of the function? Also, will this ever be called in a way that would race with the p.g assignment?

codegen/debug.go
coryb: `case "continue", "c":` is already there :)

solver/progress.go
coryb: Thanks, fixed the extra `p.g.Wait` at the end, forgot to delete it after the refactor.  As for a race, yeah maybe, I have modified it so `Sync` calls `waitNoLock` since the `Sync` already holds the lock, and then changed  `Wait` to acquire a lock then call `waitNoLock`

codegen/debug.go
coryb: I don't object to having special cases eventually, I think we just want to live with it for a bit first so we can more clearly identify what frustrates us.  Now that the debugger is "useful" with the exec functionality I think it will get more use soon.

solver/progress.go
aaronlehmann: I think this would be a bit more readable without the second `defer`:

```go
func (p *progressUI) waitNoLock() error {
	defer p.spp.cancel()
	<-p.done
	err := p.spp.wait()
	gerr := p.g.Wait()
	if err == nil {
		return gerr
	}
	return err
}
```

aaronlehmann (approved):

solver/progress.go
coryb: updated

aaronlehmann (approved):

PR #264, opened 2021-08-02
reconnect to forwarded tcp for each request

coryb:
Without this we can only handle one incoming connection on the UDS.
Subsequent requests try to reuse the previous connection and typically
hang.

cc @slushie @aaronlehmann This should fix the issue as per the discussion today.

### DIFF
--- a/codegen/builtin_option.go
+++ b/codegen/builtin_option.go
@@ -575,9 +575,13 @@ func (f Forward) Call(ctx context.Context, cln *client.Client, ret Register, opt
 		}
 		id = digest.FromString(localPath).String()
 	} else {
-		conn, err := net.Dial(src.Scheme, src.Host)
-		if err != nil {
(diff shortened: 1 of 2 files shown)


aaronlehmann (approved):

aaronlehmann (commented):
LGTM

codegen/builtin_option.go
```diff
@@ -586,14 +590,15 @@ func (f Forward) Call(ctx context.Context, cln *client.Client, ret Register, opt
 		}
 
 		localPath = filepath.Join(dir, "proxy.sock")
-		id = digest.FromString(localPath).String()
+		id = digest.FromString(src.String()).String()
```
coryb, line 593: This was causing cache invalidation every time we used the `forward` with `tcp` since the proxy.sock file was always in a new tmp directory.

PR #266, opened 2021-08-03
add `host` instructions to gateway exec for debugging into container

coryb:
updates to use PR branch for buildkit with this feature. Will update after PR https://github.com/moby/buildkit/pull/2294 is merged.


codegen/debug.go
```diff
@@ -598,6 +599,11 @@ func ExecWithFS(ctx context.Context, cln *client.Client, fs Filesystem, r io.Rea
 			securityMode = o.SecurityMode
 		case llbutil.NetworkOption:
 			netMode = o.NetMode
+		case llbutil.HostOption:
```
aaronlehmann, line 602: Let's also add the equivalent to `ExecWithSolveErr`. It looks like `ExtraHosts` is exposed under `Meta` in `ExecOp`.

coryb: nice catch,  forgot about that one. Fixed.

aaronlehmann (approved):

PR #269, opened 2021-08-22
fix race to set sourcemap in FileBuffer

coryb:
when running test with -race it seems there were multiple
check-then-write operations in parallel for the FileBuffer which
causes a `DATA RACE` exception

### DIFF
--- a/pkg/filebuffer/filebuffer.go
+++ b/pkg/filebuffer/filebuffer.go
@@ -37,16 +37,16 @@ func (s *Sources) Set(filename string, fb *FileBuffer) {
 
 type FileBuffer struct {
 	filename  string
-	buf       *bytes.Buffer
+	buf       bytes.Buffer
 	offset    int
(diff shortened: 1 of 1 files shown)


pkg/filebuffer/filebuffer.go
```diff
@@ -41,12 +41,14 @@ type FileBuffer struct {
 	offset    int
 	offsets   []int
 	sourceMap *llb.SourceMap
+	mu        *sync.Mutex
```
aaronlehmann, line 44: nit: This doesn't need to be a pointer.

I'm actually not sure why `buf` is a pointer - that also seems unnecessary.

aaronlehmann (approved):

pkg/filebuffer/filebuffer.go
coryb: cool, removed the ptr for mu and buf

PR #263, opened 2021-07-31
Update github.com/creachadair/jrpc2 to v0.19.0.

creachadair:
In preparation for committing to a stable v1 release for jrpc2 package, I have
made a number of updates to the library to simplify the API a bit. These
changes are discussed in creachadair/jrpc2#46 if you would like to offer input.

This commit updates the package version used here to the latest tag as of this
writing, v0.19.0, and updates the existing use to match the changes. These
changes should not produce any functional differences.

PR #271, opened 2021-09-10
Update jrpc2 to v0.24.2.

creachadair:
A follow-up to #263. No usage changes were required.


creachadair:
I believe the CI failure is unrelated to this change—the error log reports "no space left on device". All the tests pass locally, both with and without the race detector enabled.

PR #272, opened 2021-10-06
add option::image platform to specify desired platform

coryb:
Also adds a --platform option to `hlb run` to alter the default (linux/amd64) platform for images.

### DIFF
--- a/codegen/resolver.go
+++ b/codegen/resolver.go
@@ -17,13 +17,19 @@ import (
 func NewCachedImageResolver(cln *client.Client) llb.ImageMetaResolver {
 	return &cachedImageResolver{
 		cln:   cln,
-		cache: make(map[string]*imageConfig),
+		cache: make(map[cacheKey]*imageConfig),
 	}
+}
+
+type cacheKey struct {
+	ref  string
+	os   string
+	arch string
 }
 
 type cachedImageResolver struct {
 	cln   *client.Client
-	cache map[string]*imageConfig
+	cache map[cacheKey]*imageConfig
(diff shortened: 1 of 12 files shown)


codegen/resolver.go
```diff
@@ -33,8 +33,12 @@ type imageConfig struct {
 }
 
 func (r *cachedImageResolver) ResolveImageConfig(ctx context.Context, ref string, opt llb.ResolveImageConfigOpt) (dgst digest.Digest, config []byte, err error) {
+	key := ref
+	if opt.Platform != nil {
+		key += ":" + opt.Platform.OS + ":" + opt.Platform.Architecture
```
aaronlehmann, line 38: Let's use a struct as the map key.

hinshun:
Does a non-existent platform for a specific image on a registry correctly source maps the error to the `platform` function?

codegen/value.go
```diff
@@ -201,7 +202,14 @@ func (v *fsValue) Filesystem() (Filesystem, error) {
 }
 
 func (v *fsValue) Request() (solver.Request, error) {
-	def, err := v.fs.State.Marshal(context.Background(), llb.LinuxAmd64)
+	// TODO this should probably derive from DefaultPlatform(ctx) but we
+	// dont have ctx in scope here. Not sure of the implications of adding
+	// ctx to the API yet.
```
aaronlehmann, line 207: Would it be simpler to add a `defaultPlatform` field to `fsValue` and populate it when we create one a `fsValue`?

PR #275, opened 2021-12-08
capture platform for fs's so we can ensure correct execution context

coryb:
This fixes the platform that is auto-magically swapping to the "default platform" when capturing fs's from mounts.   We now preserve the requested platform from the image and plumb it through were necessary.

Also `--debug` will now have the correct platform from the fs's image.

### DIFF
--- a/codegen/builtin_fs.go
+++ b/codegen/builtin_fs.go
@@ -61,16 +61,16 @@ type Image struct{}
 func (i Image) Call(ctx context.Context, cln *client.Client, ret Register, opts Option, ref string) error {
 	var imageOpts []llb.ImageOption
 	platform := DefaultPlatform(ctx)
-	resolveOpt := llb.ResolveImageConfigOpt{
-		Platform: &platform,
-	}
 	for _, opt := range opts {
 		switch o := opt.(type) {
 		case llb.ImageOption:
 			imageOpts = append(imageOpts, o)
 		case *specs.Platform:
-			resolveOpt.Platform = o
+			platform = *o
 		}
+	}
+	resolveOpt := llb.ResolveImageConfigOpt{
+		Platform: &platform,
 	}
 	for _, opt := range SourceMap(ctx) {
 		imageOpts = append(imageOpts, opt)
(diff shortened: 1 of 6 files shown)


codegen/builtin_fs.go
```diff
@@ -69,6 +69,7 @@ func (i Image) Call(ctx context.Context, cln *client.Client, ret Register, opts
 		case llb.ImageOption:
 			imageOpts = append(imageOpts, o)
 		case *specs.Platform:
+			platform = *o
```
aaronlehmann, line 72: Nit: move the definition of `resolveOpt` down so we don't need the line below?

codegen/value.go
```diff
@@ -138,8 +139,9 @@ func (v *zeroValue) Kind() parser.Kind {
 
 func (v *zeroValue) Filesystem() (Filesystem, error) {
 	return Filesystem{
-		State: llb.Scratch(),
-		Image: &solver.ImageSpec{},
+		State:    llb.Scratch(),
+		Image:    &solver.ImageSpec{},
+		Platform: specs.Platform{OS: runtime.GOOS, Architecture: runtime.GOARCH},
```
aaronlehmann, line 144: Not sure if `ZeroValue` is used for anything in practice, but if it is, I'm wondering if this may cause some problems down the line. We may be trying to build an image for a different platform than the host.

It looks like `scratch` doesn't involve this code path, so it's probably not an issue... Makes sense to try it and see.

aaronlehmann (approved):

codegen/builtin_fs.go
coryb: fixed

aaronlehmann:
LGTM

codegen/value.go
coryb: Yeah, I don't think the ZeroValue with FileSystem is practically used.  Also in order for this Platform to be used, it would have to be the only `fs` in context, which is also probably not practically ever used.  I could leave this as an empty `specs.Platform{}` if you think that makes more sense, perhaps it would be better to have an empty Platform spec than the default one in the case we accidentally use this `Platform` ...

coryb: I have updated to leave this Platform empty, I dont see how it could be used, probably better to get an error in the future if we actually try to this empty Platform in the future.

aaronlehmann: LGTM

PR #277, opened 2021-12-08
fix zero-value fs platform

coryb:
As per usual, I was wrong: https://github.com/openllb/hlb/pull/275#discussion_r764482749

The empty Platform for the `zeroValue`  Filesystem is  used when we are operating directly on scratch, something like this:
```
fs default() {
	copy build "/output" "."
        # ^^^^ we are solving this with the platform from the `zeroValue` which now fails
	download "."
}

fs build() {
	image "alpine"
	mkdir "/output" 0o755
	run "touch /output/i-was-here"
}
```

We need to default the `zeroValue` platform to a valid value, otherwise we will get a `no match for platform in manifest: not found` error.

### DIFF
--- a/codegen/value.go
+++ b/codegen/value.go
@@ -39,14 +39,20 @@ type Register interface {
 
 type register struct {
 	Value
+	ctor func(iface interface{}) (Value, error)
 }
 
-func NewRegister() Register {
-	return &register{Value: ZeroValue()}
+func NewRegister(ctx context.Context) Register {
+	return &register{
+		Value: ZeroValue(ctx),
+		ctor: func(iface interface{}) (Value, error) {
+			return NewValue(ctx, iface)
+		},
+	}
 }
 
 func (r *register) Set(iface interface{}) (err error) {
-	r.Value, err = NewValue(iface)
+	r.Value, err = r.ctor(iface)
 	return err
 }
 
@@ -60,14 +66,14 @@ type Value interface {
 	Reflect(reflect.Type) (reflect.Value, error)
 }
 
-func NewValue(iface interface{}) (Value, error) {
+func NewValue(ctx context.Context, iface interface{}) (Value, error) {
 	switch v := iface.(type) {
 	case Value:
 		return v, nil
 	case Filesystem:
 		return &fsValue{&nilValue{}, v}, validateState(v.State)
 	case llb.State:
-		zero := ZeroValue()
+		zero := ZeroValue(ctx)
 		fs, err := zero.Filesystem()
 		if err != nil {
 			return zero, err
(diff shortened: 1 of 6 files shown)


codegen/value.go
```diff
@@ -138,8 +139,9 @@ func (v *zeroValue) Kind() parser.Kind {
 
 func (v *zeroValue) Filesystem() (Filesystem, error) {
 	return Filesystem{
-		State: llb.Scratch(),
-		Image: &solver.ImageSpec{},
+		State:    llb.Scratch(),
+		Image:    &solver.ImageSpec{},
+		Platform: specs.Platform{OS: "linux", Architecture: runtime.GOARCH},
 	}, nil
```
aaronlehmann, line 145: So it's okay if this doesn't match the arch we're building for?

coryb: Yeah, I have been thinking about this, not sure if this will cause problems, but it does seem wrong.  I suspect if we just pass in an empty `llb.Constraint{}` for the solves it will work when we dont have a platform defined.  I will double check this.

codegen/value.go
```diff
@@ -39,14 +39,20 @@ type Register interface {
...
+	ctor func(iface interface{}) (Value, error)
 }
 
-func NewRegister() Register {
-	return &register{Value: ZeroValue()}
+func NewRegister(ctx context.Context) Register {
+	return &register{
+		Value: ZeroValue(ctx),
+		ctor: func(iface interface{}) (Value, error) {
+			return NewValue(ctx, iface)
```
aaronlehmann, line 49: Preserving the context long-term like this feels a little icky. It might be better to expose a `NewValueWithPlatform` or add functional arguments to `NewValue` so we don't have to persist the context. But I think this is fine for now.

aaronlehmann (approved):

codegen/value.go
coryb: So I am wrong again.  The `zeroValue` Platform is setting the platform context for the solve, effectively becoming the default platform.  If we leave out the hlb platform constraint, eventually buildkit will default to `runtime.GOOS/runtime.GOARCH`.  This will chain down to the `image "alpine"` solve, which will use the `zeroValue` Platform.  So if we leave the platform blank then we try to solve `alpine` with `darwin/amd64` which fails.  If we hardcode the platform to be `linux/runtime.GOARCH` it generally works, unless you also use `--platform linux/arm64` to change the default platform, in that case the `zeroValue` would still be the current os arch, rather than desired default platform and effectively the `--platform` argument would be ignored.

In order to fix this properly, we need to plumb the `ctx` to everywhere `ZeroValue` is called so we assign the default platform correctly.

codegen/value.go
coryb: yeah, I was trying to avoid changing the `(*Value).Set(interface{}))` signature, which is called in ~billion places.

Issue #278, opened 2022-01-20
platform confused when using multiple archs in same build graph


coryb:
I would expect hlb to run:
```
fs default() {
	image "alpine" with platform("linux", "amd64")
	run "test $(uname -m) = x86_64" with option {
		mount fs {
			image "alpine" with platform("linux", "arm64")
			run "test $(uname -m) = aarch64" with ignoreCache
		} "/foobar" with readonly
	}
}
```
It currently doesnt, the inner `image` is having the amd64 platfrom applied from the parent scope and ignoring the platform argument.  So this graph ends up running amd64 for both vertexes.

hinshun:
When marshalling llb, you have to specify constraints for platform otherwise you get default of `runtime.GOOS` and `runtime.GOARCH`. Either way, one platform is specified, and seems to override all child platform constraints. So this is a buildkit issue, not a hlb one.

coryb:
We recently changed HLB to explictly set os/arch for all solves instead of allowing GOOS/GOARCH to be applied, I suspect the bug is there.

PR #279, opened 2022-01-20
Refactor solve tree to output branches under mount and add test for multi platform llb

hinshun:
Fixes: #278 

- Allows multi-platform images in one llb.State
- Refactors solver treeprint to separate file, and expose helper function for `codegen.Filesystem` type

I was trying to debug #278 and the treeprint output was very confusing.

Before:
```
.
└── [exec]  echo Amd
    ├── [cwd]  /
    ├── [mount]  docker-image://docker.io/library/alpine:latest => / [type=BIND]
    ├── [mount]  scratch => /foobar [type=BIND]
    ├── solve options
    └── [exec]  echo Arm
        ├── [cwd]  /
        ├── [mount]  docker-image://docker.io/library/alpine:latest => / [type=BIND]
        └── solve options
```

Note how it's just an output error but it says `scratch => /foobar` and then a sibling `[exec]` that has the actual fs.

After:
```
.
└── [exec]  echo Amd
    ├── [cwd]  /
    ├── [mount]  / [type=BIND]
    │   └── [source]  identifier:\"docker-image://docker.io/library/alpine:latest\"
    │       └── [platform]  linux,amd64
    ├── [mount]  /foobar [type=BIND]
    │   └── [exec]  echo Arm
    │       ├── [cwd]  /
    │       ├── [mount]  / [type=BIND]
    │       │   └── [source]  identifier:\"docker-image://docker.io/library/alpine:latest\"
    │       │       └── [platform]  linux,arm64
    │       └── [platform]  linux,arm64
    └── [platform]  linux,amd64
```


codegen/builtin_fs.go
```diff
@@ -70,6 +70,8 @@ func (i Image) Call(ctx context.Context, cln *client.Client, ret Register, opts
 			platform = *o
 		}
 	}
+	imageOpts = append(imageOpts, llb.Platform(platform))
```
coryb, line 73: nice, not sure how I missed this one.

coryb (approved):

PR #280, opened 2022-01-26
Refactor usage of llb.Local name and key hint

hinshun:
- Removes `solver.LocalPathDescriptionKey`, it was used only to see what local's digest actually pointed to in the solver tree. Since the path is now part of the local name, this isn't necessary anymore. (https://github.com/openllb/hlb/pull/97#issuecomment-614268142)
- Use `localPath` as name for `llb.Local`, but under the hood use abs path for sync dir
- Use our deterministic ID in `llb.SharedKeyHint` instead, not visible to end user but gives us the same benefits
- Include filename in `localPath` when specifying a file directory with `local "filename"`. Doesn't have any functional difference but makes the output better

Before
```
~/tmp
❯ cat foo.hlb
fs default() {
	local "foo.hlb"
}

~/tmp
❯ hlb run foo.hlb
[+] Building 0.0s (1/1) FINISHED
 => local://sha256:408dc57624f1359d2eaa3b6ef97dcff3b16d16125db02684613a33a8aec2bec0 (foo.hlb)
 => => transferring sha256:408dc57624f1359d2eaa3b6ef97dcff3b16d16125db02684613a33a8aec2bec0: 28B
```

After
```
~/tmp
❯ cat foo.hlb
fs default() {
	local "foo.hlb"
}

~/tmp
❯ hlb run foo.hlb
[+] Building 0.0s (1/1) FINISHED
 => local://foo.hlb (foo.hlb)
 => => transferring foo.hlb: 28B
```

### DIFF
--- a/codegen/builtin_fs.go
+++ b/codegen/builtin_fs.go
@@ -173,28 +173,31 @@ func (l Local) Call(ctx context.Context, cln *client.Client, ret Register, opts 
 		localOpts = append(localOpts, opt)
 	}
 
+	localDir := localPath
 	if !fi.IsDir() {
 		filename := filepath.Base(localPath)
-		localPath = filepath.Dir(localPath)
+		localDir = filepath.Dir(localPath)
 
 		// When localPath is a filename instead of a directory, include and exclude
 		// patterns should be ignored.
 		localOpts = append(localOpts, llb.IncludePatterns([]string{filename}), llb.ExcludePatterns([]string{}))
 	}
 
-	cwd, err := local.Cwd(ctx)
+	absPath := localPath
+	if !filepath.IsAbs(absPath) {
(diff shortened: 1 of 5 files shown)


codegen/builtin_fs.go
```diff
@@ -186,28 +187,26 @@ func (l Local) Call(ctx context.Context, cln *client.Client, ret Register, opts
 	if err != nil {
 		return err
 	}
+	absPath := filepath.Join(cwd, localPath)
```
coryb, line 190: Is there any chance that `localPath` is already a full path?

hinshun: You're right. Thanks

coryb (approved):
LGTM

aaronlehmann (approved):

PR #282, opened 2022-02-04
Use docker engine embedded buildkit by default

### LINKED ISSUE #220: speed up docker image loading into local daemon
I'm using hlb with a local buildkitd container started with: `docker run -d --name buildkitd --privileged openllb/buildkit:experimental` and building local images with the dockerLoad instruction.

One of the slowest parts of rebuilding when nothing has changed is the "sending tarball" part of the "exporting to oci image format" step:
```
 => exporting to oci image format                                                                                                                                                                                       6.7s
 => => exporting layers                                                                                                                                                                                                 0.0s
 => => exporting manifest sha256:c4db86288982ca760470b4fbafa068d5a8f5f4b746444a002b8d4483871f06e7                                                                                                                       0.0s
 => => exporting config sha256:f408dfe5501e874dd1a97fca4ecc4882c3639a18257eb4695c42fca8fa468558                                                                                                                         0.0s
 => => sending tarball                                                                                                                                                                                                  6.7s
 => => exporting manifest sha256:8cdbeb68e00e61aadd1012534d3a141393794f551092d78794f6d5caa55172ad                                                                                                                       0.0s
 => => exporting manifest sha256:2012bd6343be80e6bcaf19cf595b8969a487512b7c2c45d94f8af1b6e8da3d6c                                                                                                                       0.0s
```

docker build and buildx (https://github.com/docker/buildx/) do this much more efficiently and print out the following logs which are slightly different:
```
 => [monolith] exporting to image                                                                                                                                                                                       0.0s
 => => exporting layers                                                                                                                                                                                                 0.0s
 => => writing image sha256:81f3f327425eec9ecedf1ec21c918fd0a9c4e29e3fc881c95f99a384f1ffcda1                                                                                                                            0.0s
 => => naming to docker.io/library/monolith                                                                                                                                                            0.0s
```

It would be great to understand why buildx and docker build can do this so much faster and if it's possible for hlb to also achieve the same speed.

### LINKED ISSUE #283: Arg(...).WithError prints erroneous diagonistic
When using `return Arg(ctx, 0).WithError(...)` directly without creating spans like errors in `errdef`, an erroneous diagnostic is printed:

Example:
```sh
[+] Building 0.0s (0/0)
error: bar.hlb:3:21: docker engine does not support this exporter
 1:

aborting due to previous error
exit status 1
```

hinshun:
Fixes #220 
Fixes #283 

!!! **Need help testing on macos** !!!

Currently, the default when `BUILDKIT_HOST` is not set, is to look for a buildkit unix socket, which is the same behavior as `buildctl`. However its very rare users actually run BuildKit natively, and it's also not cross-platform.

The more common case is that users have `docker`, so by default HLB will use the buildkit embedded in docker the same way [buildx](https://github.com/docker/buildx/) does.

When using docker-engine embedded BuildKit, it shares the underlying image storage, so `dockerLoad` is many orders of magnitude faster.

Note that the exporter `moby` is only supported by embedded BuildKit:
https://github.com/moby/moby/blob/master/builder/builder-next/worker/worker.go#L221-L225


aaronlehmann:
Very cool!! This is working for me with Docker for Mac.

codegen/context.go
```diff
@@ -191,3 +194,27 @@ func DefaultPlatform(ctx context.Context) specs.Platform {
...
+func WithDockerAPI(ctx context.Context, cln dockerclient.APIClient, auth imagetools.Auth, err error, moby bool) context.Context {
+	return context.WithValue(ctx, dockerAPIKey{}, &DockerAPIClient{
+		APIClient: cln,
+		Auth:      auth,
+		Moby:      moby,
+		err:       err,
+	})
+}
+
+func DockerAPI(ctx context.Context) (*DockerAPIClient, error) {
```
aaronlehmann, line 214: Any particular reason to return a pointer here?

codegen/builtin_fs.go
```diff
@@ -708,6 +707,34 @@ func (dp DockerPush) Call(ctx context.Context, cln *client.Client, ret Register,
 		}
 	}
 
+	dockerAPI, err := DockerAPI(ctx)
+	if dockerAPI.Moby {
```
aaronlehmann, line 711: It feels a little dangerous to dereference `dockerAPI` here even if the function returned an error.

codegen/builtin_fs.go
```diff
@@ -708,6 +707,34 @@ func (dp DockerPush) Call(ctx context.Context, cln *client.Client, ret Register,
...
+		// Return error only if dockerPush is using docker engine instead of buildkit.
```
aaronlehmann, line 712: Why?

codegen/builtin_fs.go
```diff
@@ -708,6 +707,34 @@ func (dp DockerPush) Call(ctx context.Context, cln *client.Client, ret Register,
...
+		exportFS.SolveOpts = append(exportFS.SolveOpts,
+			solver.WithPushMoby(ref),
+			solver.WithCallback(func(_ context.Context, resp *client.SolveResponse) error {
+				mw := MultiWriter(ctx)
+				if mw == nil {
+					return nil
+				}
+
+				pw := mw.WithPrefix("", false)
+				return progress.Wrap(fmt.Sprintf("pushing %s", ref), pw.Write, func(l progress.SubLogger) error {
```
aaronlehmann, line 726: nit: could be simplified:

```suggestion
				return progress.Wrap("pushing "+ref, pw.Write, func(l progress.SubLogger) error {
```

codegen/builtin_fs.go
```diff
@@ -742,10 +769,81 @@ func (dp DockerPush) Call(ctx context.Context, cln *client.Client, ret Register,
...
+			}
+			if jm.Status == "Pushed" {
+				now := time.Now()
+				st.Completed = &now
+				st.Current = st.Total
+			}
+			l.SetStatus(st)
+		}
+		if jm.Error != nil {
+			parsedError = jm.Error
```
aaronlehmann, line 842: Why do we save this instead of returning it right away?

codegen/builtin_fs.go
```diff
@@ -742,10 +769,81 @@ func (dp DockerPush) Call(ctx context.Context, cln *client.Client, ret Register,
...
+				if jm.Progress != nil || jm.Status == "Pushed" {
+					now := time.Now()
+					st = &client.VertexStatus{
+						ID:      id,
+						Started: &now,
+					}
+					started[id] = st
+				} else {
+					continue
+				}
```
aaronlehmann, line 823: Why not:

```suggestion
				if jm.Progress == nil && jm.Status != "Pushed" {
					continue
				}
```

instead of if/else...

codegen/builtin_fs.go
```diff
@@ -1019,6 +1119,17 @@ func (dot DownloadDockerTarball) Call(ctx context.Context, cln *client.Client, r
 		return err
 	}
 
+	named, err := reference.ParseNormalizedNamed(ref)
+	if err != nil {
+		return errdefs.WithInvalidImageRef(err, Arg(ctx, 0), ref)
+	}
+	ref = reference.TagNameOnly(named).String()
```
aaronlehmann, line 1126: Unrelated bugfix?

hinshun: Yes sorry. I'll do it in a separate PR.

codegen/builtin_fs.go
hinshun: Tbh, I'm not too sure, I ported this code from a buildx private function. It looks like even if there is a json message error, we want to continue as long as the next decode is not also an error (io.EOF, or others).

I can experiment with exiting early but I'm not familiar with moby/moby's docker push progress code.

codegen/builtin_fs.go
hinshun: Open to another way of designing this, but basically I want hlb to function even if docker CLI is not installed. i.e. loading the docker cli could error, but if we were not using moby anyway, we don't care.

codegen/builtin_fs.go
hinshun: I guess we can also do:
```go
dockerAPI := DockerAPI(ctx)
if dockerAPI.Moby {
```

And then expose the error in `dockerAPI.Err`, what do you think?

codegen/context.go
hinshun: No reason, originally I returned `nil` in the error case.

codegen/builtin_fs.go
hinshun: Extracted to: #287 287

codegen/builtin_fs.go
coryb: Yeah `dockerApi.Err` seems more reasonable here:
```
	dockerAPI := DockerAPI(ctx)
	if dockerAPI.Moby {
		// Return error only if dockerPush is using docker engine instead of buildkit.
		if dockerAPI.Err != nil {
			return Arg(ctx, 0).WithError(err)
		}
```

codegen/builtin_fs.go
coryb: I think it so the code continues reading the `rc` stream until `EOF`.  It will return the parsedError only after `dec.Decode` has an error, which is likely `EOF`.

codegen/builtin_fs.go
```diff
@@ -1019,6 +1118,11 @@ func (dot DownloadDockerTarball) Call(ctx context.Context, cln *client.Client, r
 		return err
 	}
 
+	dockerAPI, _ := DockerAPI(ctx)
+	if dockerAPI.Moby {
+		return errdefs.WithDockerEngineUnsupported(ProgramCounter(ctx))
```
coryb, line 1123: Seems surprising that moby would not support exporting the tarball, without buildkit I think this is equivalent to `docker_ save`?

coryb (approved):
This is pretty cool!  Looks good to me, I tried out a few things on my Docker Desktop on my mac and it worked well for me.

codegen/builtin_fs.go
hinshun: Not much we can do unless we contribute to engine, I think its just a matter of no one needing it in buildx yet.

PR #284, opened 2022-02-04
Refactor RunInfo to prepare for DAP

hinshun:
- Deduplicates logic for debugger / shell-on-error on apps that embed HLB
- Currently prevents apps that embed HLB to implement things like shell on breakpoint debugger, but later will have a golang interface to programmatically control debugger
- RunInfo now takes `Stdin`, `Stdout`, and `Stderr` optionally


solver/progress.go
```diff
@@ -43,9 +43,13 @@ const (
 	LogOutputPlain
 )
 
-func WithLogOutput(con Console, logOutput LogOutput) ProgressOption {
+func WithLogOutput(stderr io.Writer, logOutput LogOutput) ProgressOption {
```
aaronlehmann, line 46: Why pass in an `io.Writer` type if it's expected to satisfy the `solver.Console` interface? Is this something that will be relaxed in the future?

cmd/hlb/command/run.go
```diff
@@ -243,14 +213,27 @@ func Run(ctx context.Context, cln *client.Client, rc io.ReadCloser, info RunInfo
...
+		opts         []codegen.CodeGenOption
+		inputSteerer *codegen.InputSteerer
+	)
+	if info.Debug {
+		pr, pw := io.Pipe()
+		r := bufio.NewReader(pr)
+		inputSteerer = codegen.NewInputSteerer(info.Stdin, pw)
+
+		debugger := codegen.NewDebugger(cln, os.Stderr, inputSteerer, r)
+		opts = append(opts, codegen.WithDebugger(debugger))
```
aaronlehmann, line 226: I'm a bit concerned about merging this before there's a mechanism for programatically controlling the debugger, since it will prevent updating apps that embed HLB and want to do things like shell-on-error.

hinshun: Sure, we can leave it dangling until everything is ready.

solver/progress.go
hinshun: If you only use plain log output, you don't need a `console.Console`. It's only because of how the `NewPrinter` function is written in buildx. We can either contribute to buildx or create an adjusted copy of the function.

hinshun: https://github.com/docker/buildx/pull/937

hinshun: Updated to use new buildx printer.

hinshun:
Closing this in favor of just maintaining the full patch in #286

PR #285, opened 2022-02-05
Decouple debugger from TUI

hinshun:
- Depends on #284
- Decouples debugger from user interaction
- Adds TUI as the first frontend for debugger
- Creates possibility of remote attach / multiple tui attaching / DAP attaching to debugger


hinshun:
Wrong branch

PR #288, opened 2022-02-05
Add support for DAP for IDE debugging

hinshun:
- Depends on #286
- Adds support for DAP
- Both codegen.Debugger and DAP shares same test suite from `codegen/debugger_testsuite.go`

vscode


.golangci.yml
```diff
@@ -13,9 +13,7 @@ linters:
     - structcheck
     - typecheck
     - unconvert
-    - unparam
```
hinshun, line 16: This was too annoying.

rpc/dapserver/server.go
```diff
@@ -0,0 +1,108 @@
...
+			return scanner.Err()
+		})
+	}
+
+	// f, err := os.Create("/tmp/hlb-dapserver.log")
+	// if err != nil {
+	// 	panic(err)
+	// }
+	// defer f.Close()
+	// log.SetOutput(f)
```
aaronlehmann, line 82: Remove or clean up?

rpc/dapserver/server.go
```diff
@@ -0,0 +1,108 @@
...
+			err := session.handleRequest(ctx)
+			if err != nil {
+				return err
+			}
+		}
+	})
+
+	session.sendWg.Wait()
+	err := g.Wait()
+	if !errors.Is(err, io.EOF) {
```
aaronlehmann, line 103: Very, very minor point - feel free to ignore: In these situations I like to limit the scope of the variable as much as possible, so I might do something like:

```go
if err := g.Wait(); !errors.Is(err, io.EOF) {
        return err
}
```

Similarly, the code a few lines above could be:

```go
if err := session.handleRequest(ctx); err != nil {
	return err
}
```

rpc/dapserver/session.go
```diff
@@ -0,0 +1,1058 @@
...
+	ProgressReportingCap
+)
+
+func (s *Session) handleRequest(ctx context.Context) error {
+	msg, err := dap.ReadProtocolMessage(s.rw.Reader)
+	if err != nil {
+		return err
+	}
+
+	s.sendWg.Add(1)
```
aaronlehmann, line 50: Would there ever be a situation where we get messages faster than we can handle them, and end up spawning too many goroutines here? I'm not sure this is a real concern, as I guess it might imply the sender is really misbehaving, and I'm not sure we need to harden against that.

aaronlehmann (approved):
Looks good! I didn't really see anything that raised concerns.

rpc/dapserver/session.go
hinshun: Yeah I'm not sure what DAP specification recommends here. I don't think this is necessary to address in this first PR for DAP support.

rpc/dapserver/server.go
hinshun: This is mostly for debugging purposes, I think eventually when dapserver can run on a unix socket / tcp and allow multiple clients it may be refactored from `hlb run --dap` to `hlb dapserver` and log output will be a flag to that command.

Multiple `codegen.Debugger` clients is necessary because IDE DAP doesn't provide a good way to do exec. My current idea is that the IDE integrated terminal is connected to the TUI debugger, so controlling the debugger can be done from both UI or terminal (`exec` being only available on the terminal).

PR #289, opened 2022-02-07
Emit BuiltinDecl async to avoid blocking codegen

hinshun:
- Supersedes https://github.com/openllb/hlb/pull/83

BuiltinDecls no longer have access to the `Register`, but simply just get the current `Value` and must return `(Value, error)`. This seems to be cleaner, since it doesn't really need to know about the `Register`.

`(Register).Set` still exists but only as syntactic sugar.

`(Register).SetAsync` is used explicitly when the value is only derived from a callback.

Should speed up any build using multiple images, and any future instruction that have asynchronous tasks (like codegen imports).


codegen/value.go
```diff
@@ -52,9 +54,71 @@ func NewRegister(ctx context.Context) Register {
...
+	return r.SetAsync(func(Value) (Value, error) {
+		return r.ctor(iface)
+	})
+}
+
+func (r *register) SetAsync(f func(Value) (Value, error)) error {
+	var prev chan Value
+	if len(r.rets) == 0 {
+		prev = make(chan Value, 1)
+		prev <- r.Value
```
aaronlehmann, line 67: Trying to understand this pattern. You create a buffered channel, write to it, and immediately read the value in the goroutine below?

I think some comments would help here.

I'm also not sure I understand the stack of `rets` and why it makes sense for `value()` to pop from the stack.

hinshun: So I'm essentially creating a chain of goroutines piped together with channels. For the zero case, I create a buffered channel to bootstrap the chain.

You're right, I don't actually need the stack, I simplified it to just keeping track of the last channel.

codegen/value.go
```diff
@@ -52,9 +54,71 @@ func NewRegister(ctx context.Context) Register {
...
+	} else {
+		prev = r.last
+	}
+
+	ret := make(chan Value)
+	r.last = ret
+
+	go func() {
+		defer close(prev)
```
aaronlehmann, line 76: Why do you need to close `prev`? Isn't the line below the last thing that will ever read from it?

codegen/value.go
```diff
@@ -52,9 +54,71 @@ func NewRegister(ctx context.Context) Register {
...
+		ret <- next
+	}()
+
+	return nil
+}
+
+func (r *register) value() Value {
+	if r.last != nil {
+		r.Value = <-r.last
+		close(r.last)
```
aaronlehmann, line 90: Same question here. You're done with `r.last`, so closing it doesn't change anything.

codegen/value.go
hinshun: Huh, I always thought you need to eventually close every channel in go. TIL

hinshun:
Updated and squashed

codegen/value.go
aaronlehmann: It's a common misunderstanding. Channels get garbage collected like other variables in Go. Closing it just causes reads to return the zero value immediately once the buffer is empty, and reads with the `x, ok := <-chan` syntax return false for `ok`.

PR #290, opened 2022-02-08
Refactor import modules into codegen using lazy async

hinshun:
- Previously, all `ImportDecl` in the AST are crawled and resolved before codegen even begins, even if they are not been used, and also disallows imports in imports since it was a one-pass.
- Imports are now emitted dynamically during codegen, making them (1) lazy, (2) recursive, allowing you to use imports within imports.
- Example, running `hlb run -t lint` no longer pulls in `image("openllb/go.hlb")` because its code path doesn't depend on the import even if the module does.

## Implementation notes
- Imports are evaluated before call{stmt,expr} arguments are evaluated because we need the type information for its arguments. The signatures are now embedded in the `(parser.Call{Stmt,Expr}).Signature` during `checker` for easy access.
- Codegen no longer needs to guess the arg types for `(*CodeGen).Evaluate` since it has access to signatures.
- Refactor `module.Resolver` to `codegen.Resolver`
- Refactor `module.Resolved` to `parser.Directory`
- Refactor `cmd/hlb/command/context.go` client to `client.go` for re-use in `langserver`
- Once import is resolved, changed `obj.Data = imod.Scope` to `obj.Data = imod`, since we need access to both scope and directory interface.
- `Callable` and `BuiltinDecl` were refactored around due to golang import cycles
- Import related tests moved from `checker_test` to `codegen_test`
- Linter recursive removed because its already recursive in codegen's dynamic imports
- Move lint tests to `linter_test`, updated `checker_test` and `codegen_test` so they don't expect lint transforms

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,38 +20,8 @@ import (
 // BuildKit which supports a subset of the exporters and special `moby`
 // exporter.
 func Client(ctx context.Context, addr string) (*client.Client, context.Context, error) {
-	var (
-		dockerCli *dockercommand.DockerCli
-		auth      imagetools.Auth
-		err       error
-	)
 	// Attempt to connect to a healthy docker engine.
-	for _, f := range []func() error{
-		func() error {
-			dockerCli, err = dockercommand.NewDockerCli()
-			return err
-		},
-		func() error {
-			return dockerCli.Initialize(flags.NewClientOptions())
-		},
-		func() error {
-			_, err = dockerCli.Client().ServerVersion(ctx)
-			return err
-		},
-		func() error {
-			imageopt, err := storeutil.GetImageConfig(dockerCli, nil)
-			if err != nil {
-				return err
-			}
-			auth = imageopt.Auth
-			return nil
-		},
-	} {
-		err = f()
-		if err != nil {
-			break
-		}
-	}
+	dockerCli, auth, err := NewDockerCli(ctx)
 
 	// If addr is empty, connect to BuildKit using connection helpers.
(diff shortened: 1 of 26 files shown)


module/resolve.go
```diff
@@ -234,11 +187,16 @@ type remoteResolved struct {
 	closed chan struct{}
 }
 
-func (r *remoteResolved) Digest() digest.Digest {
+func (r *remoteDirectory) Path() string {
+	// TODO: cache imports in home dir
```
hinshun, line 191: Leaving for follow up PR.

client.go
```diff
@@ -0,0 +1,71 @@
...
+// BuildKit which supports a subset of the exporters and special `moby`
+// exporter.
+func Client(ctx context.Context, addr string) (*client.Client, context.Context, error) {
+	var (
+		dockerCli *dockercommand.DockerCli
+		auth      imagetools.Auth
+		err       error
+	)
+	// Attempt to connect to a healthy docker engine.
+	for _, f := range []func() error{
```
aaronlehmann, line 29: Never seen this pattern before. Do you think it's better than the normal `if err := foo(); err != nil { return err }` approach?

hinshun: It evolved into something weird, I fixed it here: https://github.com/openllb/hlb/pull/290/commits/7296571611076ca7b8bba03c5ea3e31b724f1bbe

coryb (approved):
Hard one to review, I think I get the gist of it, nothing jumps out to me as a problem.

PR #295, opened 2022-02-11
Fast path register set when no async has been queued up

hinshun:
No need to create goroutines and channels unnecessarily

### DIFF
--- a/codegen/value.go
+++ b/codegen/value.go
@@ -55,8 +55,16 @@ func NewRegister(ctx context.Context) Register {
 }
 
 func (r *register) Set(iface interface{}) error {
+	// If there are no async queued up, fast path towards setting the register.
+	val, err := r.ctor(iface)
+	if r.last == nil {
+		if err == nil {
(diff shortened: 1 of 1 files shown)


codegen/value.go
```diff
@@ -55,8 +55,14 @@ func NewRegister(ctx context.Context) Register {
 }
 
 func (r *register) Set(iface interface{}) error {
+	val, err := r.ctor(iface)
+	if r.last == nil {
+		// If there are no async queued up, fast path towards setting the register.
+		r.value = val
```
aaronlehmann, line 61: Is it correct to set `r.value = val` even if `r.ctor` returned an error? I notice the slow path would set it to `&errorValue{err}`.

hinshun: Yes you're right, we don't need to set `&errorValue{err}` but we shouldn't override `r.value` with a nil.

aaronlehmann (approved):

PR #300, opened 2022-02-14
Refactor func signature into struct

hinshun:
- Necessary for the debugger to match the function signature as one `ast.Node` with a start -> end position
- Rename `fun` variables to be `fd` consistently across the code base


parser/ast/unparse.go
```diff
@@ -147,22 +147,28 @@ func (e *Export) Unparse(opts ...UnparseOption) string {
...
-		params = fd.Params.Unparse(opts...)
+	body := ""
+	if fd.Body != nil {
+		body = fmt.Sprintf(" %s", fd.Body.Unparse(opts...))
 	}
 
-	effects := ""
-	if fd.Effects != nil {
-		effects = fmt.Sprintf(" %s", fd.Effects.Unparse(opts...))
+	return fmt.Sprintf("%s%s", fd.Sig.Unparse(opts...), body)
```
aaronlehmann, line 155: Nit: `fd.Sig.Unparse(opts...) + body` is simpler

aaronlehmann (approved):

PR #301, opened 2022-02-14
Add parsed modules to module lookup registry in context

hinshun:
- Breaking up #286 into smaller PRs

### DIFF
--- a/parser/ast/context.go
+++ b/parser/ast/context.go
@@ -1,0 +1,43 @@
+package ast
+
+import (
+	"context"
+	"sync"
+)
+
+type modulesKey struct{}
+
+func WithModules(ctx context.Context, mods *ModuleLookup) context.Context {
+	return context.WithValue(ctx, modulesKey{}, mods)
+}
+
+func Modules(ctx context.Context) *ModuleLookup {
+	mods, ok := ctx.Value(modulesKey{}).(*ModuleLookup)
+	if !ok {
+		return NewModules()
+	}
+	return mods
+}
+
+type ModuleLookup struct {
+	mods map[string]*Module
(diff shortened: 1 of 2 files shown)


parser/ast/context.go
```diff
@@ -0,0 +1,43 @@
...
+	mu   sync.Mutex
+}
+
+func NewModules() *ModuleLookup {
+	return &ModuleLookup{
+		mods: make(map[string]*Module),
+	}
+}
+
+func (ml *ModuleLookup) Get(filename string) *Module {
```
coryb, line 33: This isn't used anywhere, right?

coryb (approved):

parser/ast/context.go
```diff
@@ -0,0 +1,43 @@
...
+	ml.mu.Lock()
```
aaronlehmann, line 34: Would it make sense to use a `RWLock` and use `RLock()` here? Or does it not really matter?

Alternatively, we could use `sync.Map`, though the docs say:

```
The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
```

aaronlehmann (approved):

parser/ast/context.go
hinshun: It's not used atm, I just wanted the eventual review of the debugger to be easier.

parser/ast/context.go
hinshun: I can switch to `sync.RWMutex` to keep things simple.

hinshun: Done.

PR #306, opened 2022-02-21
Evaluate arguments asynchronously and delay register to value un-lazying

hinshun:
This makes codegen fully asynchronous, as arguments will also be evaluating asynchronously.

```hlb
fs default() {
    image "alpine"
    copy image("other") "/etc/os-release" "/a"
}
```

Previously in https://github.com/openllb/hlb/pull/289, `SetAsync` was introduced and each block literal was executed asynchronously by creating a chain of goroutines piping to each other. However in the example above, `image("other")` will only start resolving image config once `image "alpine"` was done because each line depended on the line above it.

In this PR, `image("other")` will be executed asynchronously, and `copy` will block until its arguments have resolved.

If a builtin `Callable` required an argument like `input Value` instead of `input Filesystem`, it is further delayed until the builtin decides to resolve `input` to a `Filesystem`, allowing the builtin to do things asynchronously before consuming `input`. (Used in a the `fs cache(fs input, string ref)` experiment). This is implemented by `lazyValue`.

### DIFF
--- a/codegen/value.go
+++ b/codegen/value.go
@@ -36,17 +36,19 @@ type Option []interface{}
 type Register interface {
 	Value() Value
 	Set(interface{}) error
-	SetAsync(func(Value) (Value, error)) error
+	SetAsync(func(Value) (Value, error))
 }
 
 type register struct {
+	debug bool
 	value Value
-	last  chan Value
+	last  Value
 	ctor  func(iface interface{}) (Value, error)
 }
 
 func NewRegister(ctx context.Context) Register {
 	return &register{
+		debug: GetDebugger(ctx) != nil,
 		value: ZeroValue(ctx),
 		ctor: func(iface interface{}) (Value, error) {
 			return NewValue(ctx, iface)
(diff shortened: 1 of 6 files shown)


codegen/value.go
```diff
@@ -208,6 +217,55 @@ func (v *errorValue) Reflect(t reflect.Type) (reflect.Value, error) {
...
+	once  *sync.Once
+	done  chan struct{}
+	val   Value
+}
+
+func (v *lazyValue) wait() {
+	v.once.Do(func() {
+		v.val = <-v.valCh
+		close(v.done)
+	})
```
aaronlehmann, line 231: Another way to do this:

```suggestion
	val, ok := <-v.valCh:
	if ok {
		v.val = val
		close(v.valCh)
	}
```

Then you don't need `once` or `done`.

aaronlehmann (approved):

codegen/value.go
hinshun: If there are multiple goroutines waiting on `<-v.valCh` is it possible a context switch happens after line 228, and calls `(*lazyValue).Filesystem` for example, causing a nil pointer dereference before line 230 `v.val = val` is assigned?

aaronlehmann: That is not possible. Only one goroutine will receive a value from the channel (`ok == true`). The rest will block until the `close` on line 231, and receive `ok == false`.

hinshun: Ah makes sense.

PR #307, opened 2022-02-21
Add support for stargz images

hinshun:
See: https://github.com/moby/buildkit/blob/master/docs/stargz-estargz.md

Example manifest:
```json
{
   "mediaType": "application/vnd.oci.image.manifest.v1+json",
   "schemaVersion": 2,
   "config": {
      "mediaType": "application/vnd.oci.image.config.v1+json",
      "digest": "sha256:cba6a63f22c24075f2f0c9efeea9603dc9845aa3afcec9cd061c08f5d12ecc75",
      "size": 1031
   },
   "layers": [
      {
         "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
         "digest": "sha256:25af8f56fba7da444a7be3ff6a6df9ad9c15b19c624e1c2338bb0d1b1b226f69",
         "size": 314668980
      },
      {
         "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
         "digest": "sha256:7c084638d9c588cc1efc79c73a44e1424899eb10b4067119393a0dfab8400876",
         "size": 485,
         "annotations": {
            "containerd.io/snapshot/stargz/toc.digest": "sha256:c859b4593016fd4de87ba9dd20076c800428dc2f4b3b8b5e2b3ba1dc5d25f6d3",
            "io.containers.estargz.uncompressed-size": "4096"
         }
      }
   ]
}
```

### DIFF
--- a/builtin/lookup.go
+++ b/builtin/lookup.go
@@ -265,6 +265,14 @@ var (
 					},
 				},
 			},
+			"option::dockerPush": {
+				Func: map[string]FuncLookup{
+					"stargz": {
+						Params:  []*ast.Field{},
+						Effects: []*ast.Field{},
+					},
+				},
+			},
 			"option::frontend": {
 				Func: map[string]FuncLookup{
 					"input": {
@@ -1184,6 +1192,16 @@ fs diff(fs base)
 # expanded the same as the docker CLI.
 # @return an option to push the filesystem to a registry.
 fs dockerPush(string ref) binds (string digest)
+
+# Compress the image as a eStargz image before pushing.
+#
+# eStargz is a lazily pullable image format. It is compatible to OCI/Docker
+# images so it can be pushed to standard container registries and runnable on
+# eStargz-agnostic runtimes like Docker.
+# See: https://github.com/containerd/stargz-snapshotter
+#
+# @return an option to compress image as eStargz before pushing.
+option::dockerPush stargz()
 
 # Loads the filesystem as a Docker image to the docker client found in your
 # environment.
--- a/codegen/builtin_fs.go
+++ b/codegen/builtin_fs.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/containerd/containerd/platforms"
+	"github.com/containerd/containerd/remotes/docker"
 	"github.com/docker/buildx/util/imagetools"
 	"github.com/docker/buildx/util/progress"
 	"github.com/docker/distribution/reference"
@@ -28,6 +30,7 @@ import (
 	"github.com/openllb/hlb/local"
 	"github.com/openllb/hlb/parser"
 	"github.com/openllb/hlb/pkg/llbutil"
+	"github.com/openllb/hlb/pkg/stargzutil"
 	"github.com/openllb/hlb/solver"
 	"github.com/pkg/errors"
 	fstypes "github.com/tonistiigi/fsutil/types"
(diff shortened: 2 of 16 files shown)


builtin/lookup.go
```diff
@@ -1185,6 +1193,16 @@ fs diff(fs base)
...
 
+# Compress the image as a eStargz image before pushing.
+#
+# eStargz is a lazily pullable image format. It is compatible to OCI/Docker
+# images so it can be pushed to standard container registries and runnable on
+# eStargz-agnostic runtimes like Docker.
+# See: https://github.com/containerd/stargz-snapshotter
+#
+# @return an option to compress image as eStargz before pushing.
+option::dockerPush stargz()
```
aaronlehmann, line 1204: Does it make sense to call the option `stargz` if we are using the `estargz` format? Not really familiar with the terminology, so maybe people use `stargz` to refer to the whole class of formats, in which case this seems okay.

hinshun: It's like saying `bpf` instead of `ebpf`. The snapshotter itself is called `stargz-snapshotter` so I opted for `stargz`... nevertheless I don't mind either way.

aaronlehmann (approved):

hinshun:
Pending discussion in: https://github.com/moby/buildkit/issues/2663

hinshun:
BuildKit's stargz compression default behavior is to leave standard layers alone, creating a mixed mode image where some layers are standard and some are stargz. I believe this is intuitive for users, so in the spirit of HLB, we should always end up with a fully stargz image, and have options available for mixed mode later.

See: https://github.com/moby/buildkit/issues/2663#issuecomment-1047300669

hinshun:
Also it seems that gateway exec unlazy stargz image, looking at our gateway code, nothing sticks out on our usage.

See: https://github.com/moby/buildkit/issues/2664

codegen/builtin_fs.go
```diff
@@ -787,6 +817,75 @@ func (dp DockerPush) Call(ctx context.Context, cln *client.Client, val Value, op
...
+// hasNonStargzLayer traverses a manifest by resolving the reference and
+// walking over its children to see if there's any non-stargz layers.
+//
+// If ref points to a manifest list, the platform matcher is used to only
+// consider the current platform. Although BuildKit supports building manifest
+// lists, and multi-platform conversion is technically possible client side,
+// it requires pulling blobs locally (not via BuildKit) which is undesirable.
+//
+// See: https://github.com/containerd/stargz-snapshotter/blob/v0.6.4/nativeconverter/estargz/estargz.go
+func hasNonStargzLayer(ctx context.Context, resolver remotes.Resolver, platform platforms.MatchComparer, ref string) (bool, error) {
```
aaronlehmann, line 829: Can we add a test for this?

aaronlehmann (approved):

codegen/builtin_fs.go
hinshun: Okay, and maybe this makes sense to move to this to under `pkg/`

hinshun:
Added tests and disabled for moby engine until it supports stargz.

codegen/builtin_fs.go
```diff
@@ -91,13 +91,27 @@ func (i Image) Call(ctx context.Context, cln *client.Client, val Value, opts Opt
...
-		resolver = ImageResolver(ctx)
+		st         = llb.Image(ref, imageOpts...)
+		image      = &solver.ImageSpec{}
+		resolver   = ImageResolver(ctx)
+		resolveOpt = llb.ResolveImageConfigOpt{Platform: &platform}
 	)
 
+	// For some reason, llb.ResolveModeDefault defaults to
+	// llb.ResolveModeForcePull on BuildKit but it defaults to
+	// llb.ResolveModePreferLocal on docker engine.
```
aaronlehmann, line 102: Any reason not to be explicit and unconditionally set `resolveOpt.ResolveMode = llb.ResolveModeForcePull.String()` regardless of what we expect the default to be?

aaronlehmann (approved):

PR #308, opened 2022-02-22
Add cache builtin

hinshun:
Experimental. Maybe redundant with buildkit's export cache options.


language/builtin.hlb
```diff
@@ -3,6 +3,14 @@
 # @return a scratch filesystem.
 fs scratch()
 
+# Caches an input fs by its vertex digest.
+# If the image exists then use that image instead of executing input.
+#
+# @param input a filesystem to cache by vertex digest.
+# @param ref a docker registry reference. if not fully qualified, it will be
```
aaronlehmann, line 10: Missing phrase at the end?

language/builtin.hlb
```diff
@@ -3,6 +3,14 @@
 # @return a scratch filesystem.
 fs scratch()
 
+# Caches an input fs by its vertex digest.
+# If the image exists then use that image instead of executing input.
```
aaronlehmann, line 7: We should probably note that this pushes the image as well as building it.

PR #312, opened 2022-02-24
Add support for running modules from git scheme uri

hinshun:
Supports new schemes for module uris: 
- `file://` or empty scheme are for local files
- `git+https://` use https
- `git+ssh://` use ssh
- `git://` auto

On `git://` and `git+ssh://` mode, it relies on env `$SSH_AUTH_SOCK`, dials the agent and tests the SSH connection. If it fails on `git+ssh://`, then it fails `hlb run`. If it fails on `git://` then it falls back to `git+https://`.

Known hosts from `~/.ssh/known_hosts` are passed to `llb.KnownSSHHosts(keys string)`.

```sh
❯ hlb run -t golang git://github.com/openllb/hlb:/go.hlb
[+] Building 4.2s (3/3) FINISHED
 => CACHED [module git://github.com/openllb/hlb:/go.hlb] git://github.com/openllb/hlb                                          3.0s
 => resolve image config for docker.io/library/golang:1.17.5-alpine                                                            1.0s
 => CACHED docker-image://docker.io/library/golang:1.17.5-alpine                                                               0.1s
 => => resolve docker.io/library/golang:1.17.5-alpine                                                                          0.1s
 ```
 
 When running with either new scheme, it treats `local` relative to the `ast.Directory` the module is in. This means relative to the ref returned from gateway solve. This allows you to run remote hlb modules without cloning the repo.
 
 Other commands also work with the new uri:
 ```sh
 ❯ hlb mod tree git://github.com/openllb/hlb
[+] Building 5.5s (2/2) FINISHED
 => resolve image config for docker.io/openllb/go.hlb:latest                                                                   4.9s
 => [import go] docker-image://docker.io/openllb/go.hlb:latest                                                                 0.5s
 => => resolve docker.io/openllb/go.hlb:latest                                                                                 0.2s
 => => sha256:f05b0316512e2b5fce823ed50c11990ab8448b9eef9d0644f490c0f77d5a782e 1.95kB / 1.95kB                                 0.3s
 => => extracting sha256:f05b0316512e2b5fce823ed50c11990ab8448b9eef9d0644f490c0f77d5a782e                                      0.0s
git@github.com/openllb/hlb/build.hlb
├── [go]  /go.hlb
│   └── [go]  sha256:910cae9/module.hlb
└── [mkdocs]  /mkdocs.hlb
    └── [go]  sha256:910cae9/module.hlb
```

Including debugging:
```sh
❯ hlb run --debug -t golang git://github.com/openllb/hlb:/go.hlb
#1 [module git://github.com/openllb/hlb:/go.hlb] git://github.com/openllb/hlb
#1 1.400 ref: refs/heads/master	HEAD
#1 1.515 0bbdeac4d615b51d5991f92588dd291bfb376085	HEAD
#1 2.789 0bbdeac4d615b51d5991f92588dd291bfb376085	refs/heads/master
#1 CACHED
Type help for a list of commands
(hlb) s
git@github.com/openllb/hlb/go.hlb:13:4:
9  │ string versionCmd() {
10 │ 	"git describe --match 'v[0-9]*' --tags --dirty='.dirty' --always | sed 's/^v//'"
11 │ }
12 │
13 │ fs golang() {
   │    ^^^^^^
14 │ 	image "golang:1.17.5-alpine"
15 │ }
16 │
17 │ fs build(fs src, string package, string verPackage) {
(hlb)
```


builtin/lookup.go
```diff
@@ -769,12 +763,6 @@ option::local includePatterns(variadic string pattern)
 # @return an option to sync files that don&#39;t match any pattern.
 option::local excludePatterns(variadic string pattern)
 
-# Sync the targets of symlinks if path is to a symlink.
-#
-# @param path a list of paths to files that may be symlinks.
-# @return an option to sync the targets of symlinks.
-option::local followPaths(variadic string path)
```
hinshun, line 776: `llb.FollowPaths` unfortunately doesn't behave the same as copy's `(llb.CopyInfo).FollowSymlinks`. Very similar, but not identical.

If upstream can converge these two, we can support this again.

aaronlehmann: I'm curious what the difference is.

hinshun:
@aaronlehmann Also currently this PR doesn't utilize `llb.KnownSSHHosts` so its not secure atm.

Earthly for example, loads user's known hosts:
https://github.com/earthly/earthly/blob/main/buildcontext/gitlookup.go#L589-L595

And has some hard coded keyscans for popular hosts (github.com, bitbucket.com):
https://github.com/earthly/earthly/blob/main/buildcontext/gitlookup.go#L60-L62

builtin/lookup.go
hinshun: For `llb.FollowPaths`, it uses `FollowLinks` from here:
https://github.com/tonistiigi/fsutil/blob/master/walker.go#L49

For `(llb.CopyInfo).FollowSymlinks`, it uses `rootPath` from here (`ResolveWildcards` ends up using `rootPath` too):
https://github.com/tonistiigi/fsutil/blob/master/copy/copy.go#L110

which ends up using `RootPath` from `containerd/continuity`:
https://github.com/containerd/continuity/blob/main/fs/path.go#L225-L227

cmd/hlb/command/run.go
```diff
@@ -303,11 +308,83 @@ func ParseModuleURI(ctx context.Context, cln *client.Client, stdin io.Reader, ur
 		}
 		defer f.Close()
 		return parser.Parse(ctx, f)
+	case "git", "ssh":
```
aaronlehmann, line 311: Should we use `gitssh://` or something like that instead of `git://`? It could be confusing that `ssh://` implies a Git repo, and in the future we may want to support remote access via SSH.

aaronlehmann: Also, do we want to support `http://` and `https://` protocol schemes to access Git repos? It seems those are different from `git://`

https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols

cmd/hlb/command/run.go
```diff
@@ -303,11 +308,83 @@ func ParseModuleURI(ctx context.Context, cln *client.Client, stdin io.Reader, ur
...
+	var (
+		sessionOpts []llbutil.SessionOption
+		remote      string
+		root        string
+	)
+
+	sockPath := local.Env(ctx, "SSH_AUTH_SOCK")
+	if sockPath == "" {
+		root = u.Host + u.Path
+		remote = "https://" + root
```
aaronlehmann, line 336: Confused about why this uses `https://`. Wouldn't the user-specified URL be either a `git://` or `ssh://` URL?

cmd/hlb/command/run.go
```diff
@@ -303,11 +308,83 @@ func ParseModuleURI(ctx context.Context, cln *client.Client, stdin io.Reader, ur
...
+	} else {
+		// Use ssh protocol.
+		user := u.User
+		if user == "" {
+			user = "git"
+		}
+		root = user + "@" + u.Host + u.Path
+		remote = "ssh://" + root
```
aaronlehmann, line 344: Also confused about how this can force `ssh://`. How do we know that the remote Git repo is available via SSH?

codegen/builtin_fs.go
```diff
@@ -191,7 +192,36 @@ func (l Local) Call(ctx context.Context, cln *client.Client, val Value, opts Opt
 
 		// When localPath is a filename instead of a directory, include and exclude
 		// patterns should be ignored.
-		localOpts = append(localOpts, llb.IncludePatterns([]string{filename}), llb.ExcludePatterns([]string{}))
+		localOpts = []llb.LocalOption{
```
aaronlehmann, line 195: Why is it okay to overwrite `localOpts` here?

codegen/builtin_fs.go
```diff
@@ -191,7 +192,36 @@ func (l Local) Call(ctx context.Context, cln *client.Client, val Value, opts Opt
...
+		}
+
+		fs := Filesystem{
+			State: llb.Scratch().File(
+				llb.Copy(llb.NewState(defop), localPath, "/", copyOpts...),
+				SourceMap(ctx)...,
+			),
+			Platform: DefaultPlatform(ctx),
+		}
+		return NewValue(ctx, fs)
```
aaronlehmann, line 224: Can you explain this additional copy?

codegen/builtin_fs.go
hinshun: `llb.IncludePatterns` and `llb.ExcludePatterns` overrides under the hood, not appends.

The change is equivalent, just more explicit.

cmd/hlb/command/run.go
hinshun: Ideally, we should test first and fallback to https, you are right.

cmd/hlb/command/run.go
hinshun: This is HLB's git scheme, not the one specified by `git`. The idea is to provide a URI that can also specify branch and file path within a repo, and have logic to detect whether to use https vs ssh.

Other schemes could be something like `dockerfile://`, or like `gist://`. I didn't want the `https://` to arbitrarily mean git repo, but you're right in that `ssh://` breaks the pattern too.

cmd/hlb/command/run.go
hinshun: Yeah `gitssh://` maybe makes more sense here.

cmd/hlb/command/run.go
aaronlehmann: I think ideally we would respect the scheme the user provided. `git clone git://host/repo` means something different from `git clone https://host/repo` or `git clone ssh://host/repo`.

codegen/builtin_fs.go
aaronlehmann: It seems weird that we'd ignore the options provided in the `opts` arg to this function and `SourceMap` when `localPath` is not a dir. Does that just happen to be the correct behavior for all the options we currently support? It seems very subtle.

cmd/hlb/command/run.go
hinshun: I looked around for [other conventions](https://gist.github.com/edolstra/40da6e3a4d4ee8fd019395365e0772e7#flake-references), after reading around, I came up with this, WDYT?
- `git+https`: use https
- `git+ssh`: use ssh
- `git`: auto

aaronlehmann: SGTM. Only thing I'm wondering is if we need a way to force the classic "git" protocol. `git+git://` would definitely be weird. I'm not sure this is used much anymore, so maybe we shouldn't specifically handle it for now.

cmd/hlb/command/run.go
hinshun: The native `git://` from git itself doesn't seem like a great choice: https://gist.github.com/grawity/4392747#protocols-to-choose-from-when-cloning

codegen/builtin_fs.go
hinshun: You're right, it's missing SourceMap. I'll change this back to append.

hinshun:
Updated description of PR.
- Tests SSH agent connection
- Added loading of `~/.ssh/known_hosts` and use of `llb.KnownSSHHosts`

parser/util.go
```diff
@@ -19,6 +22,30 @@ func ResolvePath(root, path string) (string, error) {
...
+
+	if path[0] != '~' {
+		return path, nil
+	}
+
+	if len(path) > 1 && path[1] != '/' && path[1] != '\\' {
+		return "", errors.New("cannot expand user-specific home dir")
+	}
+
+	// Non-cgo required, available since go1.12
```
aaronlehmann, line 40: Non-cgo required means this only works with non-cgo builds? Is that right?

parser/util.go
```diff
@@ -19,6 +22,30 @@ func ResolvePath(root, path string) (string, error) {
...
+func ExpandHomeDir(path string) (string, error) {
+	if len(path) == 0 {
+		return path, nil
+	}
+
+	if path[0] != '~' {
+		return path, nil
+	}
+
+	if len(path) > 1 && path[1] != '/' && path[1] != '\\' {
```
aaronlehmann, line 36: Should this check for `filepath.Separator` specifically?

uri.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	cfg := &ssh.ClientConfig{
+		User: user,
+		Auth: []ssh.AuthMethod{
+			ssh.PublicKeysCallback(sshAgent.Signers),
+		},
+		HostKeyCallback: knownHosts,
+	}
+	cln, err := ssh.Dial("tcp", host+":22", cfg)
```
aaronlehmann, line 177: We might want to add the capability to specify a custom port at some point.

uri.go
```diff
@@ -0,0 +1,201 @@
...
+	st := llb.Git(remote, u.Branch, gitOpts...)
+	def, err := st.Marshal(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var pw progress.Writer
+	mw := codegen.MultiWriter(ctx)
+	if mw != nil {
+		pw = mw.WithPrefix(fmt.Sprintf("module %s", uri), true)
```
aaronlehmann, line 136:
```suggestion
		pw = mw.WithPrefix("module "+uri, true)
```

uri.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return err
+	}
+	defer cln.Close()
+
+	return nil
+}
+
+func defaultKnownHosts() (ssh.HostKeyCallback, error) {
+	filename, err := parser.ExpandHomeDir("~/.ssh/known_hosts")
```
aaronlehmann, line 187: I think this will work correctly on a host that uses `\` separators. It might be more idiomatic to do something like `filepath.Join("~", ".ssh", "known_hosts")` though. Up to you.

uri.go
```diff
@@ -0,0 +1,201 @@
...
+			return nil, err
+		}
+		ssh = true
+	case "git": // auto
+		if sockPath != "" {
+			err = testSSHAgent(sockPath, u.Host, user)
+			if err == nil {
+				ssh = true
+			}
+			// TODO: HLB logging system.
```
aaronlehmann, line 92: Agreed, it would be good to get some visibility into this (eventually).

aaronlehmann (approved):

parser/util.go
hinshun: Yes, sorry the language is confusing. It works without cgo.

parser/util.go
hinshun: This comes from: https://github.com/mitchellh/go-homedir/blob/master/homedir.go#L55-L58
Which is already cross-platform, didn't want to change the logic too much there.

uri.go
hinshun: I think I can parse the port out. Let me add that.

hinshun: This is done.

PR #315, opened 2022-03-04
Support URIs in import decl

hinshun:
- `(*codegen.CodeGen).EmitImport` uses the same `ParseModuleURI`.

Example:
```hlb
import go from "git://github.com/openllb/hlb@master:/go.hlb"

fs default() {
	go.lint src
}

fs src() {
	local "." with includePatterns("**/*.go", "go.mod", "go.sum", ".golangci.yml", ".git", "**/*.json")
}
```

### DIFF
--- a/codegen/uri.go
+++ b/codegen/uri.go
@@ -1,10 +1,9 @@
-package hlb
+package codegen
 
 import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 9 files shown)


rpc/langserver/server.go
```diff
@@ -702,7 +722,7 @@ func (ls *LangServer) textDocumentCompletionHandler(ctx context.Context, params
 }
 
 func isPositionWithinNode(pos lsp.Position, node ast.Node) bool {
-	return ast.IsPositionWithinNode(node, pos.Line-1, pos.Character-1)
```
hinshun, line 705: This was a bug.

codegen/uri.go
```diff
@@ -34,61 +33,77 @@ var (
...
+}
+
+func parseModuleGitURI(ctx context.Context, cln *client.Client, uri string) (*ast.Module, error) {
 	u, err := gitscheme.Parse(uri)
 	if err != nil {
 		return nil, err
 	}
+	dockerAPI := DockerAPI(ctx)
+	if dockerAPI.Moby {
+		return nil, errdefs.WithDockerEngineUnsupported(ProgramCounter(ctx))
```
hinshun, line 84: We should add docker version (or embedded buildkit version if possible) to `dockerAPI` when `v0.8` finally lands in docker-ce.

aaronlehmann (approved):

PR #321, opened 2022-05-31
Temporarily disable async register setting

aaronlehmann:
We've been seeing problems that appear to be tied to the new async ref
resolution behavior. Some complex build graphs spawn thousands of
goroutines, use large amounts of memory, and get stuck contending on the
build concurrency limit. Unfortunately, we haven't been able to come up
with a repro case so far. My hunch is that maybe the asynchronous
behavior bypasses some important caching and causes a complexity
explosion (or there is some other bug causing the symptoms...).

Unless someone is interested in digging into this further, my proposal
is to disable asynchronous register setting for now to fix HLB master.
I'm planning to test this PR, and if I confirm it solves the problem, we
can consider merging it.

### DIFF
--- a/codegen/value.go
+++ b/codegen/value.go
@@ -48,7 +48,14 @@ type register struct {
 
 func NewRegister(ctx context.Context) Register {
 	return &register{
-		debug: GetDebugger(ctx) != nil,
+		// Temporarily disable async sets.
+		// TODO: Revert this once async register behavior is stable.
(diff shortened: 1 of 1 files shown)


aaronlehmann:
This didn't seem to fix the problem; I suspect something else broke in recent HLB commits that's not strictly related to setting registers asynchronously.

PR #323, opened 2022-07-20
Revert lazy-eval

coryb:
The async usage is causing too many issues for us, we have not been able to use master for a while.  It works great for small-ish graphs but where there are hundreds of vertices in the graph that are lazy evaluated we were seeing thousands of goroutines get launched which seemed to overwhelm buildkit, so we started seeing timeouts, disconnects and other issues that appear to be related to load/usage.

The concurrency limiting work helped with the load, but we ended up with deadlocks where all the running work was blocked on lazy-evals that were unable to run due to limits.

I tried several approaches to try to reproduce this in an isolated test case, but so far have been unable.

Lets revert the lazy-eval stuff for now until, hopefully we can come up with some test-cases to reproduce the issues and see if we can fix the laxy-eval stuff in the future.


aaronlehmann (approved):
LGTM, but we should probably confirm this solves the problem before merging.

I previously tried https://github.com/openllb/hlb/pull/321, which I thought would effectively force synchronous eval, but it didn't help. The lazy eval changes feel like a good place to start with tracking down the regression, but I suppose it could be something else like the buildkit client upgrade.

hinshun (approved):
:-(

coryb:
I am going to hold off on this, we still see some issues even after this revert.  The latest error we are seeing is:

aaronlehmann:
I think a key issue with the async ref resolution is that it leaks `Build` calls. `NewRemoteDirectory`'s `Build` goroutine blocks inside the `Build` until the returned object's `Close` method is called, but this never happens.

Before this refactor, refs were always resolved and used right away, then closed: https://github.com/openllb/hlb/blob/5ab9a768e2b611e539b28c903a80cbffcf088b11/module/resolve.go#L402-L406

After the refactor, the refs are long-lived and are stored in the `ast.Module` structure for future use. The call to `Close` on the resolved directory was dropped in the commit that did this refactor, and doesn't appear to be invoked anywhere (removing `Close` from the `ast.Directory` interface and removing the implementation from `remoteDirectory` doesn't break compilation). This may have been intentional based an an assumption that there will be a reasonable number of imports; however some programatically generated use cases involve very large numbers of imports, leading to many parallel sessions.

This also explains why the concurrency limiter is causing problems. Because `NewRemoteDirectory` builds are never allowed to complete, we hit the concurrency limit and get stuck at that point.

I am not sure what the right fix is. One option is to go ahead with this revert, and that's probably the easiest path. I'm not sure if there's a place where resolved directories can safely be closed under the new design. If not, one approach would be to always close refs immediately after resolving them, but save the arguments we used for the resolution, so we can re-resolve later if needed. Again, I'm not sure if it's worth doing at this moment.

@hinshun Do you have any thoughts on this?

hinshun:
I think it makes sense to revert the changes until the lifecycle of these refs can be better constrained in the next iteration of async.

coryb:
I am going to close this one,  after digging in a bit more we also need to revert several more PRs all the way back to #290 which is pretty painful with merge conflicts that are pretty difficult to resolve.  The new idea is to reimplement NewRemoteDirectory so we can manage the lifecycle of the solve (current plan is to try to move solve into the Open/Stat calls so it is still lazy evaluated).

hinshun:
@coryb Would it make sense to branch off from before #290 and work from there instead of reverting everything? Cherry-pick what you need, etc (or even name it as master, while keeping current master as another branch).

coryb:
yeah, I started there,  reset to 290 and started cherry picking, but it also just became a ton of merge conflicts very quickly.

I think the only real problem we are having is the remote directory lingering solve that never closes, so I think we can probably "fix forward" by simplifying that for now, it will be suboptimal but maybe viable.  Will do some testing...

PR #324, opened 2022-07-21
Return cached SourceMap in FileBuffer

aaronlehmann:
openllb/hlb#286 changed `FileBuffer` to return a newly constructed `SourceMap` on each call to the `SourceMap` method. This causes a dramatic increase in the gRPC request size, because It turns out that BuildKit deduplicates source maps based on pointer address: https://github.com/moby/buildkit/blob/d21254e7f74b49a84c3fbe1b13260cb23954cf92/client/llb/sourcemap.go#L57

Reuse the same `SourceMap` structure when nothing has changed.

### DIFF
--- a/pkg/filebuffer/filebuffer.go
+++ b/pkg/filebuffer/filebuffer.go
@@ -63,12 +63,13 @@ func (b *BufferLookup) All() []*FileBuffer {
 }
 
 type FileBuffer struct {
-	filename string
-	buf      bytes.Buffer
-	offset   int
-	offsets  []int
-	mu       sync.Mutex
-	onDisk   bool
+	filename  string
+	buf       bytes.Buffer
+	offset    int
+	offsets   []int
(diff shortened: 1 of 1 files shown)


pkg/filebuffer/filebuffer.go
```diff
@@ -113,6 +120,11 @@ func (fb *FileBuffer) Bytes() []byte {
 }
 
 func (fb *FileBuffer) Write(p []byte) (n int, err error) {
+	fb.mu.Lock()
+	defer fb.mu.Unlock()
+
+	fb.sourceMap = nil
```
aaronlehmann, line 126: I suspect this isn't necessary, but added it out of an abundance of caution since I wasn't sure why `SourceMap` was changed to stop caching the `sourceMap` in the first place.

coryb (approved):

hinshun (approved):

PR #327, opened 2022-08-01
Support extra env vars for the debug shell

yun-wang:
The `env []string` support was removed from `ExecWithSolveErr` in #286. Adding it back because it's required in newt.

This PR adds `env []string` to the client-facing `Exec` method, instead of the underlying implementation (`ExecWithSolveErr` or `ExecWithFS`).

### DIFF
--- a/codegen/gateway_exec.go
+++ b/codegen/gateway_exec.go
@@ -19,7 +19,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-func ExecWithFS(ctx context.Context, cln *client.Client, fs Filesystem, opts Option, stdin io.Reader, stdout, stderr io.Writer, args ...string) error {
+func ExecWithFS(ctx context.Context, cln *client.Client, fs Filesystem, opts Option, stdin io.Reader, stdout, stderr io.Writer, extraEnv []string, args ...string) error {
 	var (
 		securityMode pb.SecurityMode
 		netMode      pb.NetMode
@@ -33,8 +33,9 @@ func ExecWithFS(ctx context.Context, cln *client.Client, fs Filesystem, opts Opt
 		cwd = fs.Image.Config.WorkingDir
 	}
 
-	env := make([]string, len(fs.Image.Config.Env))
+	env := make([]string, len(fs.Image.Config.Env), len(fs.Image.Config.Env)+len(extraEnv))
 	copy(env, fs.Image.Config.Env)
+	env = append(env, extraEnv...)
 
 	user := fs.Image.Config.User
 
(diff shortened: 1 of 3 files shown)


codegen/gateway_exec.go
```diff
@@ -33,8 +33,7 @@ func ExecWithFS(ctx context.Context, cln *client.Client, fs Filesystem, opts Opt
 		cwd = fs.Image.Config.WorkingDir
 	}
 
-	env := make([]string, len(fs.Image.Config.Env))
-	copy(env, fs.Image.Config.Env)
+	env = append(env, fs.Image.Config.Env...)
```
aaronlehmann, line 36: It might be a good idea to make a copy of `env` before appending to it. It may not be safe to append to a slice that's passed in and might be reused for other purposes.

codegen/gateway_exec.go
```diff
@@ -288,7 +287,7 @@ func ExecWithSolveErr(ctx context.Context, c gateway.Client, se *solvererrdefs.S
 		Args:         args,
 		Cwd:          exec.Meta.Cwd,
 		User:         exec.Meta.User,
-		Env:          exec.Meta.Env,
+		Env:          append(exec.Meta.Env, env...),
```
aaronlehmann, line 290: Same thought here about making a copy.

hinshun (approved):
Same thoughts about the copy comments from @aaronlehmann.
I also think at some point, we should probably pull stdio + env + anything else into a struct.

codegen/gateway_exec.go
```diff
@@ -33,8 +33,9 @@ func ExecWithFS(ctx context.Context, cln *client.Client, fs Filesystem, opts Opt
 		cwd = fs.Image.Config.WorkingDir
 	}
 
-	env := make([]string, len(fs.Image.Config.Env))
+	env := make([]string, len(fs.Image.Config.Env)+len(extraEnv))
```
aaronlehmann, line 36: This doesn't look right if you're using `append` below. You would want:

```go
env := make([]string, len(fs.Image.Config.Env), len(fs.Image.Config.Env)+len(extraEnv))
```

codegen/gateway_exec.go
```diff
@@ -284,11 +285,15 @@ func ExecWithSolveErr(ctx context.Context, c gateway.Client, se *solvererrdefs.S
 		return err
 	}
 
+	env := make([]string, len(exec.Meta.Env)+len(extraEnv))
```
aaronlehmann, line 288: Same issue here

codegen/gateway_exec.go
yun-wang: Oops indeed! Fixed now.

aaronlehmann (approved):

PR #328, opened 2022-08-02
Exit early on ErrDebugExit after Solve

yun-wang:
This PR fixes an issue where the debugger would hang at [`d.mode = <-d.control`](https://github.com/openllb/hlb/blob/master/codegen/debugger.go#L369) if `d.mode` is already `DebugTerminate`, and there is no more `control` value from the channel.

### DIFF
--- a/codegen/debugger.go
+++ b/codegen/debugger.go
@@ -320,6 +320,10 @@ func (d *debugger) yield(ctx context.Context, scope *ast.Scope, node ast.Node, v
 
 		err = req.Solve(ctx, d.cln, MultiWriter(ctx))
 		if err != nil {
+			// If debugger has an error, continue to exit.
+			if d.err != nil {
+				return d.err
+			}
 			yieldErr = ProgramCounter(ctx).WithError(err)
 		}
 	}


aaronlehmann (approved):
LGTM

codegen/debugger.go
```diff
@@ -320,6 +320,10 @@ func (d *debugger) yield(ctx context.Context, scope *ast.Scope, node ast.Node, v
 
 		err = req.Solve(ctx, d.cln, MultiWriter(ctx))
 		if err != nil {
+			// If debugger has an error, continue to exit.
+			if d.err != nil {
```
yun-wang, line 324: I used `if d.err != nil` to be consistent with the check [above](https://github.com/openllb/hlb/blob/master/codegen/debugger.go#L311). I'm also okay with checking for `ErrDebugExit` explicitly here (`errors.Is(err, ErrDebugExit)`).

coryb: it seems like if we have any debugger error, we should be exiting, so it probably does not make sense to special case for ErrDebugExit?

coryb (approved):

codegen/debugger.go
yun-wang: I think the only value for `d.err` currently is `ErrDebugExit`, but I agree it's better to check for any debugger error here, if new errors get added later.

hinshun:
Hmm.. so if I'm remembering this right, this part of the code attempts to cast the return value as a `Request` and then evaluate it, so that the debugger won't continue past execution that would otherwise error out in a non-debugging `hlb run`.

We capture the solve error as a `yieldError` so that it displays in the TUI and you can remain in the debugger in order to reverse, re-solve, do other debugging tasks.

There seems to be a deadlock when the debugger enters `DebugTerminate` between lines:
https://github.com/openllb/hlb/blob/83da43256207303ec69ced2a8a7eda18c25723d0/codegen/debugger.go#L315-L321

Is that right? It's not clear to me yet this is the right fix. Perhaps there should be some kind of concurrency control around giving clients the opportunity to terminate vs continuing on to evaluating the request.

yun-wang:
Correct, there is a deadlock when the debugger enters `DebugTerminate` in the [`req.Solve`](https://github.com/openllb/hlb/blob/master/codegen/debugger.go#L321) method.

I bumped into the issue with our `ControlDebugger` implementation for `shell-on-error`. Once the debug shell exits, we call [`Terminate()`](https://github.com/openllb/hlb/blob/master/codegen/debugger.go#L234) to end the debugging session. The `ErrDebugExit` error is returned from the `req.Solve` method (from the error handler). The solve error is actually `ErrDebugExit`. It doesn't make sense to remain in the debugger in this case.

yun-wang:
Regarding the current fix, is there a scenario where `d.err != nil` after `req.Solve` returns, and we'd still want to remain in the debugger?

hinshun:
Okay, I'm happy to move forward with this PR. Though would be nice to write a test that hangs without this fix. It sounds like it can be reduced to a simple reproduction.

hinshun (approved):

PR #329, opened 2022-08-03
close session after solves

coryb:
This is now required after a fix to buildkit that prevented sessions from being shared my multiple solves.
https://github.com/moby/buildkit/pull/2995

A future optimization here might be to cache sessions for re-use within the hlb build rather than creating/closing for each solve.

This also fixes tests that are expecting amd64 as the default platform.


pkg/llbutil/session.go
```diff
@@ -63,7 +64,8 @@ func NewSession(ctx context.Context, opts ...SessionOption) (*session.Session, e
 	}
 
 	// By default, forward docker authentication through the session.
-	attachables := []session.Attachable{authprovider.NewDockerAuthProvider(os.Stderr)}
+	dockerConfig := config.LoadDefaultConfigFile(os.Stderr)
```
hinshun, line 67: We should probably plumb stderr at some point instead of using `os.Stderr`.

codegen/builtin_fs.go
```diff
@@ -252,10 +253,10 @@ func (l Local) Call(ctx context.Context, cln *client.Client, val Value, opts Opt
 	fs.SessionOpts = append(fs.SessionOpts, llbutil.WithSyncedDir(id, filesync.SyncedDir{
 		Name: localPath,
 		Dir:  localDir,
-		Map: func(_ string, st *fstypes.Stat) bool {
+		Map: func(_ string, st *fstypes.Stat) fsutil.MapResult {
 			st.Uid = 0
 			st.Gid = 0
-			return true
+			return fsutil.MapResultKeep
```
hinshun, line 259: What does this do?

solver/request.go
```diff
@@ -71,6 +71,7 @@ func (r *singleRequest) Solve(ctx context.Context, cln *client.Client, mw *Multi
 	})
 
 	g.Go(func() error {
+		defer s.Close()
```
hinshun, line 74: Were we just never closing sessions?

hinshun (approved):

codegen/builtin_fs.go
coryb: it is the same result, the fsutil walk api changed to require an enum to be resturns vs the `true` (to mean "keep going")

solver/request.go
aaronlehmann: BuildKit was automatically closing them on the first solve, which was broken: https://github.com/moby/buildkit/pull/2995

coryb: There is a PR to buildkit linked in the description, it was always closed in the client.Solve even if it was passed in externally with the intent to use as a shared session.  WIth the buildkit fix we now need to close our own sessions since we are creating them outside the client.Solve.  This does give us the flexibility in the future to actually share and reuse sessions across multiple solves though.

pkg/llbutil/session.go
coryb: yeah good idea, we should probably stash stdio on the context in the command impls so we can fetch them as necessary

PR #331, opened 2022-08-11
Add platform prefix to MultiWriter

yun-wang:
This PR adds a platform prefix to Buildkit progress output, because the output is confusing when Newt solves multi-platform builds in parallel. However, the current implementation doesn't cover all output in plain mode. Buildkit doesn't give us a way to inject prefixes to the rest of the progress output:
- https://github.com/moby/buildkit/blob/master/util/progress/progressui/printer.go#L131
- https://github.com/moby/buildkit/blob/master/util/progress/progressui/printer.go#L198

Progress output with the current implementation:

Screen Shot 2022-08-11 at 4 07 09 PM

Let me know what you think, or if you have any suggestions.

### DIFF
--- a/cmd/hlb/command/run.go
+++ b/cmd/hlb/command/run.go
@@ -137,6 +137,7 @@ type RunInfo struct {
 	Targets         []string
 	LLB             bool
 	LogOutput       string
+	LogPrefixes     []string
 	DefaultPlatform string // format: osname/osarch
 
 	Stdin  io.Reader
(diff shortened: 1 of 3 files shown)


hinshun:
I think the plain log output only prefixing the "header" is fine. They are always outputted in chunks together with the "header" so from a readability standpoint it looks good.

hinshun (approved):

cmd/hlb/command/run.go
```diff
@@ -175,18 +175,17 @@ func Run(ctx context.Context, cln *client.Client, uri string, info RunInfo) (err
...
 	ctx = local.WithOs(ctx, info.Os)
 	ctx = local.WithArch(ctx, info.Arch)
+	var progressOpts []solver.ProgressOption
 	if info.DefaultPlatform != "" {
 		platformParts := strings.SplitN(info.DefaultPlatform, "/", 2)
 		if len(platformParts) < 2 {
 			return fmt.Errorf("Invalid platform specified: %s", info.DefaultPlatform)
 		}
 		ctx = codegen.WithDefaultPlatform(ctx, specs.Platform{OS: platformParts[0], Architecture: platformParts[1]})
+		progressOpts = append(progressOpts, solver.WithLogPrefix(info.DefaultPlatform))
```
coryb, line 185: Perhaps we should allow for `LogPrefix []string` in the RunInfo so we can pass them in from our tools in case we want to do something special with the logging.  Then maybe if the `(*RunInfo).LogPrefix` is empty we automatically add the default platform to the prefix?

coryb (approved):
Looks good to me.

cmd/hlb/command/run.go
coryb: I guess an alternative to this would be to allow a `solver.Progress` to be passed into the `RunInfo` instead of just prefixes, not sure which option makes more sense.

yun-wang: I like making `LogPrefixes` explicit, updated the PR to support it.

PR #334, opened 2022-08-23
Set implicit local directory path to ""

### LINKED ISSUE #332: ModuleDir function returns incorrect result
This line of code appears to be wrong: https://github.com/openllb/hlb/blob/12055ea32a30db84c267c4d355ef198c2e2080fb/codegen/context.go#L76

`mod.Directory.Path()` is hardcoded to `.` by `parser.Parse`. Normally this `strings.TrimPrefix` does nothing, but if the filename is `./foo.hlb`, `strings.TrimPrefix` will remove the `.`, and `ModuleDir` will return `/`. This will make a `"download "./build"` get an error like:

```
ERROR: rpc error: code = Unknown desc = failed to create synctarget dest dir /build: mkdir /build: permission denied
```

I haven't been able to reproduce this with the `hlb` CLI, but passing in a filename starting with `./` into `command.Run` triggers this issue.

I'm not sure what this line of code is trying to accomplish. Should it just be removed, or is the correct fix something else?

aaronlehmann:
Previously, this was being set to ".", and would wrongly strip the first
"." from a path that begins with "." in `ModuleDir`. Setting it to the
empty string seems more correct: nothing will be stripped. The only
other use of this "root" is in `filepath.Join` calls inside `localDirectory`
methods, which will also do the right thing with an empty string.

Fixes #332

### DIFF
--- a/cmd/hlb/command/run.go
+++ b/cmd/hlb/command/run.go
@@ -116,7 +116,7 @@ func ParseModuleURI(ctx context.Context, cln *client.Client, stdin io.Reader, ur
 			Value:  "<stdin>",
 		})
 	}
-	dir := parser.NewLocalDirectory(".", "")
+	dir := parser.NewLocalDirectory("", "")
 	return codegen.ParseModuleURI(ctx, cln, dir, uri)
 }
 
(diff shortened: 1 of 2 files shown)


hinshun:
Was this tested against the edge cases I mentioned?

hinshun (approved):

PR #336, opened 2022-09-10
Remove SessionID from local op

aaronlehmann:
This ends up referring to an ID that isn't used by the actual session, so it causes problems. For example, "local" ops take at least 5 seconds because they try to wait for a session with this ID that never shows up: https://github.com/moby/buildkit/blob/abde08a5531d809a395cf648a31bca932b009af0/source/local/local.go#L95-L101

Ideally, we should set `SessionID` to the actual session ID instead of leaving it unspecified, because this relies on buildkit to iterate over the sessions and find one that works, which can be nondeterministic. But it seems like this would take a pretty significant refactor to create the session in advance, and then add `FSSyncProvider`s as we go. The change here should be a strict improvement over the current behavior.

### DIFF
--- a/codegen/builtin_fs.go
+++ b/codegen/builtin_fs.go
@@ -239,7 +239,10 @@ func (l Local) Call(ctx context.Context, cln *client.Client, val Value, opts Opt
 	if err != nil {
 		return nil, err
 	}
-	localOpts = append(localOpts, llb.SharedKeyHint(id))
+	localOpts = append(localOpts,
+		llb.SharedKeyHint(id),
+		llb.LocalUniqueID(id),
+	)
 
 	fs := Filesystem{
 		State:    llb.Local(localPath, localOpts...),
--- a/codegen/builtin_fs.go
+++ b/codegen/builtin_fs.go
@@ -241,11 +241,6 @@ func (l Local) Call(ctx context.Context, cln *client.Client, val Value, opts Opt
 	}
 	localOpts = append(localOpts, llb.SharedKeyHint(id))
 
-	sessionID := SessionID(ctx)
-	if sessionID != "" {
-		localOpts = append(localOpts, llb.SessionID(sessionID))
-	}
-
 	fs := Filesystem{
 		State:    llb.Local(localPath, localOpts...),
 		Platform: DefaultPlatform(ctx),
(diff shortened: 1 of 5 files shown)


codegen/builtin_fs.go
```diff
@@ -338,6 +333,7 @@ func (f Frontend) Call(ctx context.Context, cln *client.Client, val Value, opts
 				if err != nil {
 					return
 				}
+				fs.SessionOpts = sessionOpts
```
aaronlehmann, line 336: This is needed because the frontend solve creates LLB that may need access to a file source specified as a frontend option. For example, a dockerfile context needs the relevant sync provider. Before, this was only working by freak chance because combination of the 5 second delay and bogus session ID caused a session with this `FSSyncProvider` to be available at the same time and buildkit to find it.

codegen/codegen_test.go
```diff
@@ -57,13 +56,9 @@ func LocalState(ctx context.Context, t *testing.T, localPath string, opts ...llb
 
 	opts = append([]llb.LocalOption{
 		llb.SharedKeyHint(id),
+		llb.LocalUniqueID(id),
```
aaronlehmann, line 59: Would appreciate a sanity check of this change. I found it was needed for codegen tests to pass, because otherwise, without a session ID, BuildKit assigns a random ID to each `local`. I am not really clear on the difference between `SharedKeyHint` and `LocalUniqueID`.

coryb: Not sure, I think `LocalUniqueID` is different than `LocalID`.    We have a helper func: https://github.com/aaronlehmann/hlb/blob/304e004c45688ea5af8eea22ecd8df99e6a85dad/pkg/llbutil/id.go#L38
so maybe it makes sense to use that like we do [here](https://github.com/aaronlehmann/hlb/blob/304e004c45688ea5af8eea22ecd8df99e6a85dad/pkg/llbutil/id.go#L18-L22)?

aaronlehmann:
> We have a helper func:
> https://github.com/aaronlehmann/hlb/blob/304e004c45688ea5af8eea22ecd8df99e6a85dad/pkg/llbutil/id.go#L38
>
> so maybe it makes sense to use that like we do here
> <https://github.com/aaronlehmann/hlb/blob/304e004c45688ea5af8eea22ecd8df99e6a85dad/pkg/llbutil/id.go#L18-L22>
> ?
>
Yeah, the ID is derived from the output of that helper func, with the opts
hashed in.

hinshun (approved):
We should probably do it the "right" way later, but yes this seems like a strict improvement.

codegen/codegen_test.go
aaronlehmann: I chatted with Tonis and it seems like what I'm doing here is correct. Basically this serves as a unique identifier for the data when the session ID isn't available.

PR #337, opened 2022-09-13
Provide registry creds to image resolver

aaronlehmann:
Previously, HLB never provided credentials to image resolvers, so image resolution wouldn't work against registries that require authentication. This shims standard buildkit authprovider to provide credentials based on the local Docker configuration, so we reuse that logic instead of copy/pasting it.


aaronlehmann:
@hinshun: Do you know what happened with CI?

hinshun:
@coryb owns the droplet for the GH runner.

aaronlehmann:
CI passed (not sure why it was delayed)

PR #338, opened 2022-10-21
Include last modified time in LocalUniqueID

aaronlehmann:
I ran into a situation where a "local" mount was wrongly deduped with a "local" from a concurrent solve that was created by another hlb process before some files in that dir were modified.

By including the last modified timestamp in the unique ID, we an avoid this deduplication in the case where a file has actually changed.

### DIFF
--- a/pkg/llbutil/id.go
+++ b/pkg/llbutil/id.go
@@ -4,8 +4,11 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"io/fs"
 	"net"
+	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/moby/buildkit/client/llb"
 	digest "github.com/opencontainers/go-digest"
@@ -15,7 +18,7 @@ import (
 // the same content doesn't get transported multiple times when referenced
 // repeatedly.
 func LocalID(ctx context.Context, absPath string, opts ...llb.LocalOption) (string, error) {
-	uniqID, err := LocalUniqueID(absPath)
+	uniqID, err := localUniqueID(absPath)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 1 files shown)


pkg/llbutil/id.go
```diff
@@ -34,13 +37,31 @@ func LocalID(ctx context.Context, absPath string, opts ...llb.LocalOption) (stri
...
+func localUniqueID(dir string) (string, error) {
 	mac, err := FirstUpInterface()
 	if err != nil {
 		return "", err
 	}
-	return fmt.Sprintf("cwd:%s,mac:%s", cwd, mac), nil
+
+	var lastModified time.Time
+	err = filepath.Walk(dir, func(path string, info fs.FileInfo, err error) error {
+		if lastModified.IsZero() || info.ModTime().After(lastModified) {
```
coryb, line 55: This ignores the llb.LocalOption, so will potentially cause more churn than necessary if excluded files are frequently updated.  I wonder if we can use the include/exclude patterns to get more accurate here?

aaronlehmann: Hmm, I can look into using `fsutil.Walk` instead of `filepath.Walk` here (it honors include/exclude patterns).

aaronlehmann: Updated to use `fsutil.Walk` with include/exclude patterns honored. In my testing this seems to solve the original problem.

coryb: nice! it looks like the input `dir` might also be a file, so I think we can just short-circuit that scenario.

pkg/llbutil/id.go
```diff
@@ -34,13 +40,60 @@ func LocalID(ctx context.Context, absPath string, opts ...llb.LocalOption) (stri
...
+		}
+	}
+
+	fi, err := os.Stat(localPath)
+	if err != nil {
+		return "", err
+	}
+
+	lastModified := fi.ModTime()
+	if fi.IsDir() {
```
coryb, line 84: nit, we dont need to inspect the opts and/or unmarshal teh localnfo configs in the !IsDir case

coryb (approved):

hinshun (approved):

hinshun:
Nice!

PR #340, opened 2023-01-11
Upgrade to current versions of buildkit, buildx, and docker

aaronlehmann:
Some minor changes were needed for source-level compatibility with these versions.


solver/progress.go
```diff
@@ -164,25 +167,29 @@ type syncProgressPrinter struct {
...
 		w:    w,
 		out:  out,
 		mode: mode,
 	}
-	spp.reset()
-	return spp
+	if err := spp.reset(); err != nil {
+		return nil, err
+	}
+	return spp, nil
```
hinshun, line 179: Can this just be `return spp, spp.reset()`?

hinshun (approved):
LGTM, small nit

solver/progress.go
aaronlehmann: Done

PR #347, opened 2023-03-15
use ForceNoOutput for mounts that are not bound

coryb:
We dont care about mutations to mounts unless they are bound with `as name` syntax.  However, buildkit does not know this so it assumes that mounts are mutable and will may be used elsewhere, so anytime we use a mount (including `local`) it will prevent buildkit from caching the vertex.


aaronlehmann (approved):

coryb:
This was @hinshun's idea a long time ago, not sure why we didnt do it then.  It makes a great improvement on cachability when using mounts (and now we dont need to use `with readonly` all the time to just get cached results)

hinshun:
@coryb I think I didn’t fully understand the consequences of ForceNoOutput, but this makes sense. Two line diff!

coryb:
Heh, well, maybe you understood this fully ... I am not sure this is working exactly as I thought, trying to debug now.  I was able to get it working directly via llb, but still seeing poor caching with hlb, will keep looking.

